Issue #2, opened 2016-05-10
[Bug] Both Psr7Trait and JsonResponseTrait define the same $response property.


crynobone:
exception 'ErrorException' with message 'Nexmo\Entity\Psr7Trait and Nexmo\Entity\JsonResponseTrait define the same property ($response) in the composition of Nexmo\Message\Message. This might be incompatible, to improve maintainability consider using accessor methods in traits instead. Class was composed

crynobone:
![screen shot 2016-05-13 at 9 25 25 pm](https://cloud.githubusercontent.com/assets/172966/15249262/3f7b8c20-1951-11e6-97ac-48abb9445f4e.png)

tjlytle:
Good catch, the traits are a bit fluid at this point, should solidify soon. Trying to bridge compatibility with the existing APIs, and more REST-like future APIs.

PR #3, opened 2016-05-10
Remove duplicated $response property.

crynobone:
This would solve duplicated property defined by trait as reported under #2 

Signed-off-by: crynobone <crynobone@gmail.com>


tjlytle:
Thanks!

PR #14, opened 2016-09-05
Add get method for message timestamp

arubacao:
According to the docs [incoming message](https://docs.nexmo.com/messaging/sms-api/api-reference#inbound) the message timestamp is a required key in a request. Therefore i've added a corresponding method.

Furthermore I've (probably) identified a crucial bug (typo).
You've always used the key `date-received` instead of `message-timestamp`.
`date-received` is never mentioned in the official docs.

Cheers

### DIFF
--- a/src/Message/InboundMessage.php
+++ b/src/Message/InboundMessage.php
@@ -138,6 +138,11 @@ class InboundMessage implements MessageInterface, \ArrayAccess
         return $this['network'];
     }
 
+    public function getMessageTimestamp()
+    {
+        return $this['message-timestamp'];
+    }
+
     /**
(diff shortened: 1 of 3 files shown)


tjlytle:
The difference here is an inbound message webhook (which uses the `message-timestamp`) and _searching_ for that message via the API (which uses `date-received`). 

`test/Message/responses/search-inbound.json` is an actual response from the API (see [these docs](https://docs.nexmo.com/tools/developer-api/search-message)). 

On this, the only thing that's really missing is adding `getDateReceived()` (since this should be normalized to use the more common name) to `InboundMessage`. Probably best to that that through a trait both `Message` and `InboundMessage` use.

PR #17, opened 2017-01-05
Fixing abstract class method signatures in message objects

JesseObrien:
This is broken in the current `@beta` composer release. All of these classes will break (read: FATAL error) when being used because they don't implement the abstract method signature properly.

Method signature declaration here:
https://github.com/Nexmo/nexmo-php/blob/master/src/Entity/EntityInterface.php#L16

### DIFF
--- a/src/Message/Binary.php
+++ b/src/Message/Binary.php
@@ -45,9 +45,9 @@ class Binary extends Message
     /**
      * Get an array of params to use in an API request.
      */
-    public function getRequestData()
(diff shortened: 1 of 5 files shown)


tjlytle:
Thanks, this was a bit of a refactor from an earlier concept, so had not paid enough attention. Will merge, should also mention the next release will change some of the underlying request / response concepts. Hope your use was just to get it to work, and not utilizing the underlying request / response data.

PR #24, opened 2017-02-23
Changes JWT Generation Signature:

tjlytle:
This changes the signature of generateJWT. I considered allowing the last argument to be an array of claims; however, that seemed unnecessarily complex, as the 4 original arguments are really just claims.


leggetter:
`generateJwt(array $claims = [])` is the correct approach IMO.

Bump the beta version and I can't see any problems with this being a breaking change.

PR #25, opened 2017-02-25
Fixed request signature generation.

keymandll:
It now generates valid signatures in case the & and = characters are present in a parameter value.
Please note I have not updated the signature test cases.

### DIFF
--- a/src/Client/Signature.php
+++ b/src/Client/Signature.php
@@ -55,7 +55,6 @@ class Signature
 
         //create base string
         $base = '&'.urldecode(http_build_query($this->signed));
-        print $base . "\n";
 
         //append the secret
         $base .= $secret;


tjlytle:
Hey @keymandll, thanks for the PR. Can you give me an example of a parameter that failed before?

keymandll:
Please see the following example code below.

```
<?php
function test(array $params) {
    // ...
    $base = '&'.urldecode(http_build_query($params));
    return $base;
}

$tval = array('message-timestamp' => '2013-11-21 15:27:30',
              'messageId' => '020000001B0FE827',
              'msisdn' => '14843472194',
              'text' => 'Test again&hi=there',
              'timestamp' => '1385047698',
              'to' => '13239877404',
              'type' => 'text');

print test($tval);
```

It generates the output:

`&message-timestamp=2013-11-21 15:27:30&messageId=020000001B0FE827&msisdn=14843472194&text=Test again&hi=there&timestamp=1385047698&to=13239877404&type=text`

Please note the injected parameter and value "&hi=there". On the server-side "&hi=there" will simply be treated as the value of the "text" parameter thus signature generation on server-side will yield a different signature.

tjlytle:
@keymandll The referenced commit seems to be gone. At least I could not even get the patch to test locally. I've duplicated the issue and created another PR (#29) which does not change the parameters in the sent body (so you can still send a `&` in a message).

Issue #23, opened 2017-02-21
How to list account numbers?


leggetter:
Is it possible to use the client to list numbers associated with an account.

There seems to be a:

```
$client->numbers()->get($pattern);
```

But this searches for numbers based on a querystring and does appear to support list/get.

samwierema:
For reference: I've created this function now to view a listing of my numbers.
```php
public function list($queryParameters = [])
{
    $request = new Request(
        \Nexmo\Client::BASE_REST . '/account/numbers?' . http_build_query($queryParameters),
        'GET',
        'php://temp'
    );

    $response = $this->client->send($request);

    if ($response->getStatusCode() != '200') {
        throw $this->getException($response);
    }

    $body = json_decode($response->getBody()->getContents(), true);
    if (empty($body)) {
        throw new Exception\Request('number not found', 404);
    }

    foreach ($body['numbers'] as $key => $number) {
        $c = new Number();
        $c->JsonUnserialize($number);

        $body['numbers'][$key] = $c;
    }

    return $body;
}```
It's not ideal, but it works for me.

tjlytle:
@leggetter Good catch, I just need to make the `$number` argument optional.

PR #27, opened 2017-03-15
[WIP] Add numbers search, buy, cancel

marcw:
Hi,

I started implementing numbers searching, buying and cancelling. It's not yet production tested, but I'd like to know if I'm going in the right direction here.

Any comments welcome!


marcw:
This has been tested against the live API and an example has been added

tjlytle:
@marcw Will take a look, thanks!

tjlytle:
@marcw It's a great start, here are a few things to keep in mind.

**The collection methods should accept number resources as arguments.** See the ['Entity' section of the client library spec](https://github.com/Nexmo/client-library-specification/blob/master/SPECIFICATION.md#entities). This would allow the same object returned by `search()` to be passed to `buy()`.

**Throughout the library, where an entity can be used an array can also be used. ** The spec leaves this up to the language (just that library methods should accept arguments as well as entities). For the PHP library I've chosen to avoid really long argument lists in favor of arrays because:
- optional arguments get really messy as consuming code needs to duplicate default values to 'skip' them
- arrays can match the exact structure of the API request / response (an array representation of a JSON object)
- we can treat entities as arrays (using `ArrayAccess`), so a developer can ignore (for the most part) the entities entirely, act as if the client library is returning arrays and consuming arrays. 

The `search()` method is similar, in that it should accept an array (to keep the argument list manageable) or an object. In this case, the object wouldn't be a `Number` but some kind of filter. Take a look at [this call filter](https://github.com/Nexmo/nexmo-php/blob/develop-collection-refactor/src/Call/Filter.php) as an example.

Following that, if you look at what I've got going on the [`develop-collection-refactor`](https://github.com/Nexmo/nexmo-php/tree/develop-collection-refactor/src/Call) branch you can see I'm working on changing some naming. `client` to `collection` and `entity` to a specific resource (`Call` in the case of the example). The goal here is to **treat even the non-RESTful endpoints the same was as the newer RESTful endpoints**, so that means the number methods here should try to map to what we'd expect from the next version of the API. That means using methods like `post()`, `get()`, etc. And that should allow some shared code for common things. Can then alias `search()` and other current names to that.

Once you think of `search()` as a collection, there's conflict between that and the actual collection of account numbers (which is not yet supported, see #23). Since `Nexmo\Numbers\Client` should eventually become `Nexmo\Numbers\Collection`, the purchasable numbers should be its own collection. I'm not sure if that should be something like `Nexmo\Numbers\SearchCollection` or more like `Nexmo\AvailibleNumbers\Collection`. I'd like the namespacing to be similar to the API path, but I need to check if there is a solid URI structure for the next version of this API.

In the same way, there's a bit of a mismatch between a `Number` object when it's returned from the collection of numbers I own, and when it's from the collection of numbers I _can_ buy (for example, the `cost` property). So `search()` should probably not return the same object, especially so `buy()` can type hint that it expects an available number, not an own number. 

A lot of info I know, and I'm sure @leggetter would like some of the reasoning here put into a proper form in the README to help others. :smile: 

The [`develop-collection-refactor`](https://github.com/Nexmo/nexmo-php/tree/develop-collection-refactor/src/Call) should be helpful it seeing what I think is a much better way of dealing with collections and resources than what I had started with when I wrote the number code. Don't let me delay you if you want to take a look there and keep updating this PR, but if I get a chance, would you mind if I PR'd your PR with some of the changes I outline here?

marcw:
Thanks for your answers. Feel free to PR my PR! :)

lornajane:
Is this something we could pick up again from its current state? Probably a question for @mheap

mheap:
This functionality was added vis #75, #76 and #80 

Thanks for the original PR!

PR #29, opened 2017-03-27
:bug: Fix for `&` or `=` in the body:

tjlytle:
Update for #25 
- replace either with a `_` for the purpose of the signature
- keep the original arguments in the request
- TODO: need to test with inbound requests


tjlytle:
Merged.

Issue #32, opened 2017-04-18
Set callback url?


Testing989:
Hi,

I wanted to set a callback url using your library when sending a sms like this:

```
$client = new Nexmo\Client(new Nexmo\Client\Credentials\Basic(API_KEY, API_SECRET));            
$text = new \Nexmo\Message\Text(NEXMO_TO, NEXMO_FROM, 'Test message using PHP client library');
$text->setCallback($whateverurl);
```

That method is currently missing.
I added this method to the Message class and it's working:

```
    public function setCallback($callback) {
        return $this->setRequestData('callback', (string) $callback);
    }
```

Don't know if there is a reason why this method is missing or maybe the setCallback url can be done other way.

Testing989:
Still no answer?

leggetter:
@Testing989 Callback URLs are set on a per number basis. You can't set them on a per-message basis.

tjlytle:
Sorry, I think I missed this issue notification / confused it with a related one. 

@Testing989 you're right, that's an [optional parameter](https://docs.nexmo.com/messaging/sms-api/api-reference#request). You can currently pass it as the last argument in the constructor, but your code would be fine too. Feel free to send a PR, or I'll try to include the next time there's an update for the message related code.

leggetter:
You learn something every day 😊

mheap:
Closing this as it was merged in to `develop` in https://github.com/Nexmo/nexmo-php/commit/3340f2efde99ca5f97b85e70249e82488bece499

PR #34, opened 2017-05-18
Update Message.php

Testing989:
Just adding a method to add the callback url for a message as told https://github.com/Nexmo/nexmo-php/issues/32


tjlytle:
Added tests and merged into `develop`. Will be in the next release.

PR #37, opened 2017-08-11
Detect HTTP 429 when searching messages

### LINKED ISSUE #12: Wrong behavior with 429 error.
If i got 429 error, then i got response like
`<!DOCTYPE html><html><head><title>Error report</title></head><body><h1>HTTP Status 429 - message has too many concurrent requests per 1 second</h1></body></html>`.
But Client parse it as JSON and got NULL value in 
`$data = json_decode($response->getBody()->getContents(), true);`
Next Client check it and got error
`if(!$data){
            throw new Exception\Request('no message found for`' . $id . '`');
        }`
, but its wrong error, and i cant catch it by himself.
All it in Nexmo\Message\Client.

mheap:
As the endpoint does not respond with JSON on a 429, parsing the
response does not work and we incorrectly report that no message
exists. This change checks the HTTP code before attempting to inspect
the data and returns the relevant exception if needed

Resolves #12

### DIFF
--- a/src/Message/Client.php
+++ b/src/Message/Client.php
@@ -171,19 +171,20 @@ class Client implements ClientAwareInterface
         $response = $this->client->send($request);
 
         $response->getBody()->rewind();
-        
-        $data = json_decode($response->getBody()->getContents(), true);
 
-        if(!$data){
-            throw new Exception\Request('no message found for `' . $id . '`');
-        }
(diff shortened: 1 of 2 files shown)


mheap:
I think it is worth adding the ability to back off and retry, but at a later date when we standardise retry and add the ability to enable/disable it.

PR #38, opened 2017-08-11
Add ability to list account numbers

### LINKED ISSUE #23: How to list account numbers?

mheap:
When searching for numbers, I'd expect to receive a list of matching numbers rather than a single number instance. It's especially misleading at the moment as if multiple numbers match, we return a 404 and say that there are no matching numbers.

There's a more comprehensive discussion going on at https://github.com/Nexmo/nexmo-php/pull/27, but this solves the immediate issue of being able to list all numbers owned by an account by adding a new `::search` method

Resolves #23

### DIFF
--- a/src/Numbers/Client.php
+++ b/src/Numbers/Client.php
@@ -22,13 +22,13 @@ class Client implements ClientAwareInterface
     public function update($number, $id = null)
     {
         if(!is_null($id)){
-            $update = $this->get($id, true);
+            $update = $this->get($id);
         }
 
         if($number instanceof Number){
             $body = $number->getRequestData();
             if(!isset($update) AND !isset($body['country'])){
-                $data = $this->get($number->getId(), true);
+                $data = $this->get($number->getId());
                 $body['msisdn'] = $data->getId();
                 $body['country'] = $data->getCountry();
             }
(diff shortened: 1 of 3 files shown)


test/Numbers/responses/list.json
```diff
@@ -0,0 +1,30 @@
...
+        "VOICE",
+        "SMS"
+      ],
+      "voiceCallbackType": "vxml",
+      "voiceCallbackValue": "http://example.com/voice",
+      "voiceStatusCallbackUrl": "http://example.com/status"
+    },
+    {
+      "country": "US",
+      "msisdn": "12404284199",
```
aaronbassett: There are [number ranges reserved for fictitious use cases](https://en.wikipedia.org/wiki/Fictitious_telephone_number) we should probably be using numbers in these ranges in our tests to make sure we don't call people by accident 🤣

test/Numbers/responses/list.json
```diff
@@ -0,0 +1,30 @@
...
+      "country": "US",
+      "msisdn": "12404284163",
+      "moHttpUrl": "http://example.com/message",
+      "type": "mobile-lvn",
+      "features": [
+        "VOICE",
+        "SMS"
+      ],
+      "voiceCallbackType": "vxml",
+      "voiceCallbackValue": "http://example.com/voice",
```
aaronbassett: Really minor, but we should probably be using http**s** in all our examples

src/Numbers/Client.php
```diff
@@ -98,21 +116,26 @@ public function get($number)
...
-        if(!($number instanceof Number)){
-            $number = new Number();
+        // If they provided a number initially, we'll only get one response
+        // so let's reuse the object
+        if($number instanceof Number){
+            $number->jsonUnserialize($body['numbers'][0]);
+            $numbers[] = $number;
+        } else {
+            // Otherwise, we return everything that matches
+            foreach ($body['numbers'] as $n) {
```
aaronbassett: _(stylistic preference; if not common/valid in PHP please feel free to ignore)_

I tend to avoid single character variable names, most editors have tab completion now so it doesn't even save keystrokes 😏

`$number->jsonUnserialize($matchingNumber);`

is more grok-able than

`$number->jsonUnserialize($n);`

it's also a lot easier to grep for…

aaronbassett: In the same vein as the `$n` comment, I know that the variable name `$body` pre-dated your changes, but we should probably change this to something more descriptive like `$searchResults`

mheap:
@aaronbassett All good feedback! I've updated the PR with the requested changes

PR #39, opened 2017-08-11
Rename SharedSecret to SignatureSecret

### LINKED ISSUE #4: Change references of Shared Secret to Signature Secret
It's possible that the API Secret and Shared Secret could be confused. Signature Secret defines exactly what that secret is used for; for signing and for checking signatures.

We could call is request signature secret to be event clearer?

mheap:
**This is a BREAKING CHANGE**

We could alias `SignatureSecret` to `SharedSecret`, but as we're pre-1.0 that feels like something we'd just have to maintain going forwards. Provided we write a good `CHANGELOG.md` I don't see any issue with renaming this

---

This is the name that's used on the customer dashboard for this authentication method.

Resolves #4

### DIFF
--- a/src/Client/Credentials/SignatureSecret.php
+++ b/src/Client/Credentials/SignatureSecret.php
@@ -8,7 +8,7 @@
 
 namespace Nexmo\Client\Credentials;
 
-class SharedSecret extends AbstractCredentials implements CredentialsInterface
+class SignatureSecret extends AbstractCredentials implements CredentialsInterface
 {
     /**
      * Create a credential set with an API key and shared secret.
@@ -19,6 +19,6 @@ class SharedSecret extends AbstractCredentials implements CredentialsInterface
     public function __construct($key, $secret)
     {
         $this->credentials['api_key'] = $key;
-        $this->credentials['shared_secret'] = $secret;
+        $this->credentials['signature_secret'] = $secret;
     }
 }
--- a/src/Client/Credentials/SignatureSecret.php
+++ b/src/Client/Credentials/SignatureSecret.php
@@ -11,14 +11,14 @@ namespace Nexmo\Client\Credentials;
 class SignatureSecret extends AbstractCredentials implements CredentialsInterface
 {
     /**
-     * Create a credential set with an API key and shared secret.
+     * Create a credential set with an API key and signature secret.
      *
      * @param string $key    API Key
-     * @param string $secret Shared Secret
+     * @param string $signature_secret Signature Secret
      */
-    public function __construct($key, $secret)
+    public function __construct($key, $signature_secret)
     {
         $this->credentials['api_key'] = $key;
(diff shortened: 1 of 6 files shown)


tjlytle:
I suggest making it non-breaking until a 1.0 release, just because there are people using it and that milestone is a good time to break things. Also, it should be as easy as an empty class that extends the real class.

mheap:
All of these PRs are in preparation for an official 1.0 release some time in the next week or two. They'll be released as 1.0 final

aaronbassett:
Do we have a deprecation process for things like this? I like Django's process in this sort of situation

> A feature release may deprecate certain features from previous releases. If a feature is deprecated in feature release A.x, it will continue to work in all A.x versions (for all versions of x) but raise warnings. Deprecated features will be removed in the B.0 release, or B.1 for features deprecated in the last A.x feature release to ensure deprecations are done over at least 2 feature releases.

So, in this case, we would include the alias for v1 with a deprecation notice, and then remove it in the next minor version release.

aaronbassett:
@leggetter see comment above, is this something we have already? Or something that we need?

Also @judy2k this conversation will likely be relevant to you and the Python library

leggetter:
Yeah, only make breaking changes in major versions.

I'm not sure if the following is necessary:

> or B.1 for features deprecated in the last A.x feature release to ensure deprecations are done over at least 2 feature releases.

We don't have the policy anywhere. But maybe we could add it to the [client library spec](https://github.com/Nexmo/client-library-specification).

mheap:
@leggetter Technically we're going from 0.x to 1.0, so this is a major version change.

Deprecating in 1.0 and removing in 1.1 is something that we could do and would be familiar to PHP developers (see the mcrypt removal from 7.1 to 7.2 for example http://php.net/manual/en/migration71.deprecated.php)

mheap:
@leggetter Searched for `secret` and `shared` across the codebase and updated anything that was found to use `signature_secret` for clarity

src/Client/Credentials/SignatureSecret.php
```diff
@@ -19,6 +19,6 @@ class SharedSecret extends AbstractCredentials implements CredentialsInterface
     public function __construct($key, $secret)
```
leggetter: Change `$secret` to `$signature_secret` here?

test/ClientTest.php
```diff
@@ -46,10 +46,10 @@ public function setUp()
     {
         $this->http         = $this->getMockHttp();
         $this->request      = $this->getRequest();
-        $this->sharedsecret = new Client\Credentials\SharedSecret($this->api_key, $this->secret);
+        $this->signature_secret = new Client\Credentials\SignatureSecret($this->api_key, $this->secret);
```
leggetter: Still uses `$this->secret`

mheap:
@leggetter Refactored the client test to add clarity to each of the class parameters

PR #68, opened 2017-08-23
Add helper methods for GET/POST/PUT/DELETE

mheap:
These are available for use in the situation that Nexmo add new API
endpoints, but the customer cannot upgrade their library. By reusing
the existing `Client::send` method, we add authentication etc for free.

The desired way to implement this would be for consumers to create
their own `Zend\Diactoros\Request` (or anything that is
`Psr\Http\Message\RequestInterface` compatible) objects to pass to
`Client::send`. This option is still available - these new methods are
for those that want a quick way to get moving.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -22,6 +22,7 @@ use Nexmo\Entity\EntityInterface;
 use Nexmo\Verify\Verification;
 use Psr\Http\Message\RequestInterface;
 use Zend\Diactoros\Uri;
+use Zend\Diactoros\Request;
 
 /**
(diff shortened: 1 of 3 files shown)


mheap:
@leggetter JWTs are added automatically in the `send` method if the client has been configured with a `Keypair` auth https://github.com/Nexmo/nexmo-php/blob/9f632cdecd1fe6cb99fd6c6f6c0de3fd575c9d1e/src/Client.php#L300

mheap:
Spoke to @leggetter and this is fine for now. If we end up using these methods internally they'll need updating with header support etc

PR #73, opened 2017-09-05
Fix user agent string to conform to the client library spec

mheap:
See https://github.com/Nexmo/client-library-specification/blob/master/SPECIFICATION.md#reporting

Instead of having multiple parts delimited with slashes, each
name/version pair must be it's own entry, separated with a space.

This commit also allows the user to specify an app/name version to help
us identify their traffic if required

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -38,7 +38,7 @@ use Zend\Diactoros\Request;
  */
 class Client
 {
-    const VERSION = '1.0.0-beta1';
+    const VERSION = '1.0.0';
 
     const BASE_API  = 'https://api.nexmo.com';
     const BASE_REST = 'https://rest.nexmo.com';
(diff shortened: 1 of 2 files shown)


leggetter:
Should we standardize on passing an app name and app version in to the constructor within the spec? Node does a simple userAgent append. See https://github.com/Nexmo/nexmo-node/blob/master/src/Nexmo.js#L56

May be good to aim for consistency. But not a big deal.

mheap:
@leggetter Looking at nexmo-node, the behaviour is the same. If they don't provide any UA we don't do anything. If they do provide a UA we append it to the end.

The only difference is that the node library trusts whatever the user provides (e.g. `a/b/c` which could break our parsing engine) whereas the PHP lib asks for name/version separately and performs validation

leggetter:
I agree the PHP library solution is nicer. My question is: should we standardise on this and make the PHP library the first to adhere to the new standard? Or should we leave this detail to the library developer?

cc @judy2k

mheap:
Standardising sounds like a good idea to me. Especially around disallowed characters + other validation

judy2k:
I missed these comments. I think standardising makes sense, but as a recommendation. I'll make a note to add it to the doc.

PR #76, opened 2017-09-06
Support for purchasing a number

### LINKED ISSUE #55: Implement `buyNumber`

mheap:
At the moment we can't distinguish why a purchase has failed, so we
just pass the server error back to the user.

When a number is successfully purchased, there's no useful information
in the response. We could perform a `GET` on the number and return an
object but it feels a little heavy handed.

For now, the method doesn't return anything, and throws exceptions on
any errors. If the method call succeeds, we purchased the number
successfully

Resolves #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ API Coverage
     * [ ] Top Up
     * [X] Numbers
         * [X] Search
-        * [ ] Buy
+        * [X] Buy
         * [ ] Cancel
         * [X] Update
 * Number Insight
--- a/src/Numbers/Client.php
+++ b/src/Numbers/Client.php
@@ -136,6 +136,49 @@ class Client implements ClientAwareInterface
         return $numbers;
     }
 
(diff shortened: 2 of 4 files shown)


mheap:
@arubacao You expressed an interest in this functionality. Fancy reviewing the pull request?

arubacao:
@mheap Sure.

That's very unfortunate that this endpoint doesn't provide better responses - both the error and success side. 
But I have a suggestion to make:
If it can be assumed that a `200` is a successful purchase, the function could return a (unfetched) `Number` instance.
The `Number` [constructor](https://github.com/Nexmo/nexmo-php/blob/develop/src/Numbers/Number.php#L42) only needs the number and country. 
On the other hand, it could be confusing because the `Number` is incomplete. Maybe a second `get` request isn't a bad idea.

mheap:
@arubacao Yeah, we couldn't decide internally either. I've left it without the extra `get` as the user can always perform one if required

PR #85, opened 2017-10-17
Support 204 response from PUT /call in addition to 200

mheap:
This is related to a voice API bug. The endpoint may revert to
returning data in future with a 200 response, so we need to support
both HTTP codes simultaneously

### DIFF
--- a/src/Call/Call.php
+++ b/src/Call/Call.php
@@ -113,7 +113,8 @@ class Call implements EntityInterface, \JsonSerializable, JsonUnserializableInte
         $request->getBody()->write(json_encode($payload));
         $response = $this->client->send($request);
 
-        if($response->getStatusCode() != '200'){
+        $responseCode = $response->getStatusCode();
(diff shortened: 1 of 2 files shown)


src/Call/Call.php
```diff
@@ -113,7 +113,8 @@ public function put($payload)
         $request->getBody()->write(json_encode($payload));
         $response = $this->client->send($request);
 
-        if($response->getStatusCode() != '200'){
```
aaronbassett: The [JIRA ticket](https://nexmoinc.atlassian.net/browse/VAPI-1118) isn't public, but we should probably add some explanation as a comment as to why we now accept `200` & `204` here

mheap: That info is in the commit message rather than the code. Want me to move it over?

aaronbassett: I'd think to do a `git blame`, if you think that's enough I'm happy for this to be merged. Your choice :+1:

PR #88, opened 2017-11-10
Add ability to specify additional parameters for searchOwned

### LINKED ISSUE #87: Missing params for `account/numbers` search
Besides the number, one cannot define any other params like `size`.

[Method `searchOwned($number = null)`](https://github.com/Nexmo/nexmo-php/blob/e03dc4c0e71075c348f194955b6ed01e0d833ada/src/Numbers/Client.php#L127)

[API docs: list-owned-numbers](https://developer.nexmo.com/api/developer/numbers#list-owned-numbers)

mheap:
There is a white list of allowed parameters. Any other parameters
will throw an exception

Resolves #87

### DIFF
--- a/src/Numbers/Client.php
+++ b/src/Numbers/Client.php
@@ -124,9 +124,9 @@ class Client implements ClientAwareInterface
         return $this->handleNumberSearchResult($response, null);
     }
 
-    public function searchOwned($number = null)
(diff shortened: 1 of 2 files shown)


mheap:
@arubacao This is released as https://github.com/Nexmo/nexmo-php/releases/tag/1.1.2 now

PR #94, opened 2017-12-19
Add support for the conversion endpoint

### LINKED ISSUE #43: Implement `submitConversion`

mheap:
Resolves #43

### DIFF
--- a/src/Conversion/Client.php
+++ b/src/Conversion/Client.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Nexmo\Conversion;
+
+use Nexmo\Client\ClientAwareInterface;
+use Nexmo\Client\ClientAwareTrait;
+use Nexmo\Client\Exception;
+
+
+class Client implements ClientAwareInterface
+{
+    use ClientAwareTrait;
+
+    public function sms($message_id, $delivered, $timestamp=null)
+    {
+        return $this->sendConversion('sms', $message_id, $delivered, $timestamp);
+    }
+
+    public function voice($message_id, $delivered, $timestamp=null)
+    {
+        return $this->sendConversion('voice', $message_id, $delivered, $timestamp);
+    }
+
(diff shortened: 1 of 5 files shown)


src/Conversion/Client.php
```diff
@@ -0,0 +1,63 @@
...
+        if ($timestamp) {
+            $params['timestamp'] = $timestamp;
+        }
+
+        $response = $this->client->postUrlEncoded(
+            \Nexmo\Client::BASE_API . '/conversions/'.$type.'?'.http_build_query($params),
+            []
+        );
+
+        if($response->getStatusCode() != '200'){
```
mheap: We only check the HTTP code for this API

src/Conversion/Client.php
```diff
@@ -0,0 +1,63 @@
...
+            throw $this->getException($response);
+        }
+    }
+
+    protected function getException(ResponseInterface $response)
+    {
+        $body = json_decode($response->getBody()->getContents(), true);
+        $status = $response->getStatusCode();
+
+        if($status === 402) {
```
mheap: 402 is special - it means conversions aren't activated on your account

src/Conversion/Client.php
```diff
@@ -0,0 +1,63 @@
...
+            $e = new Exception\Request("This endpoint may need activating on your account. Please email support@nexmo.com for more information", $status);
```
afolson: Do we standardize these exception texts across all of the libraries anywhere? No worries if not, just thinking out loud here.

mheap: This is the first support controlled one that I've added. Eventually I'd like to centralise `getException` but the focus is on coverage to start, refactoring second

arubacao: @mheap To be honest, I'd suggest a refactoring (or complete rebuild) first. I'm pretty confident that the custom nexmo exceptions are never actually thrown. Guzzle throws a ClientException long before you can even check the http statuscode.

PR #96, opened 2017-12-19
Automatically grab version from composer.json

### LINKED ISSUE #91: Read Client::VERSION from composer.json
We currently have to remember to update `Client::VERSION` on each release. We could use https://github.com/Ocramius/PackageVersions to do it dynamically instead

mheap:
Resolves #91


lornajane:
This is an excellent feature and was basically ready to go - it'll help us track the library metrics since the version number will always be right (I spent ten minutes earlier trying to understand why a 1.7 install was reporting as 1.2 - of course the constant is out of date). However it needs a *major version bump* to current stable PHP 7.1. Do we want to do a major version bump on the PHP lib version number?

(also the travis builds are passing, but not reporting back to github, I've just restarted them)

lornajane:
@mheap I can't add you as a reviewer because it is your PR! Please let me know what you think though

lornajane:
Note that we're holding this to get a couple of other things on the 1.x branch before we merge this since it requires a major version bump for the move to PHP 7.

mheap:
@lornajane Now that 1.8.0 has gone out I think we can merge this and release as 1.9.0

PR #102, opened 2018-03-12
Add getFinalStatus method

ursuleacv:
The key "status" is not always available in the response and it throws a notice error "Undefined index: status" when calling `$message->getStatus();`

Also added `getFinalStatus()` method


mheap:
Thanks for the contribution @ursuleacv!

I'll tag a release this week with this fix in there

PR #106, opened 2018-04-28
Add ability to set base URL

mheap:


arubacao:
Why should you customize the `baseurl`? Is the sandbox coming back?

mheap:
Afraid not @arubacao. Customising the base URL is useful when debugging, pointing to services such as hookbin

mheap:
It looks like the fix was cherry picked / is no longer in this PR after a rebase. Updated the title to reflect the actual changes (set base URL)

lornajane:
I rebased this against master to check there were no new uses of the constants needing replacement. Also I've separated this so that we don't set both URLs to the same all the time - now we can override either or both, as needed. I'm currently working with the Acount API which uses both endpoints and I might well want to test that they are correctly different, so I thought I'd just add two parameters rather than the one which felt like a bit of a shortcut.

PR #109, opened 2018-05-30
Add Redact API

mheap:


mheap:
@afolson I added codeclimate with the intention of refactoring in the future, so I think we can ignore the warnings for now

PR #112, opened 2018-07-16
Add the current framework to the user agent when making requests

mheap:


lornajane:
I'm really interested in the approach that parses `composer.json` - I don't want to block this change as I think it's valuable though.  Why do we have to wait for PHP before we can parse the `composer.json`?  I'd be happy to start adding PHP 7 features with a check for version at this point, if the rest of the team feels OK about that.

mheap:
We already have the framework name - `getLaravelVersion()` returns `laravel/x.y.z`

I'll add slim v2 and v3 and update the PR

The library for parsing composer files is PHP 7 only (https://github.com/Ocramius/PackageVersions), which is our blocker (unless we want to add IO by reading composer.json to the library, which I'd be hesitant about)

mheap:
I'd like to finish this, but it's on hold due to time constraints at the moment. If anyone wants to pick it up, feel free

dragonmantank:
Closing this. For reference, we've talked internally about this a few times, but honestly this isn't something we should be trying to determine - there are too many frameworks out there to try and look for their individual signatures.

What we _will_ be doing is offering packages in most of the major frameworks that help with bootstrapping the client, and in those cases we'll mark the user agent appropriately.

PR #123, opened 2018-09-25
Fix typo: exsisting -> existing

eduherminio:


eduherminio:
Please feel free to commit that in my branch in the way you consider more appropriate.

PR #130, opened 2018-10-10
HTTPS is better :)

m1guelpf:


mheap:
Hey @m1guelpf, 

As this is part of the error message returned, I'd rather keep it as HTTP. It's not as secure, but it allows people to copy their entire error message and search for it on this page. Updating to HTTPS would prevent them from doing this.

Thanks for the PR, and good luck for the rest of hacktoberfest!

PR #133, opened 2018-10-13
Remove dead code in getException()

gy741:
Hello.

I do not think the ```$application``` parameter is needed.

Because it is not used

what do you think?

Thanks.


lornajane:
Sorry, I just merged another PR that removed it in the call, but I'd like to leave it in place in the function signature for compatibility with other methods.

PR #137, opened 2018-10-16
Use sames characters from documentation

roukmoute:
We have few problems about carriage return.
And we have compared with your [form](https://developer.nexmo.com/messaging/sms/guides/concatenation-and-encoding#ConcatenationComponent).
And we have a big difference with/without carriage return.
Without it, it should take 2 messages and with it, it should take 3 messages for this kind of message:

"Bonjour Totoro,
Votre colis OT 123456789 est en cours de livraison.
Celui-ci est prévu pour une heure d'arrivée estimée le October 15, 2018 17:00.
Plus d'informations sur https://foo.bar/baz"

It works not all the time, but sometimes we do not receive the sms.

Here it is an example:

```php
$client = new Client(new Basic(KEY, SECRET));
$text = new Text(
    '+336xxx,
    $sms->from(),
    "Bonjour Totoro,
Votre colis OT 123456789 est en cours de livraison.
Celui-ci est prévu pour une heure d'arrivée estimée le October 15, 2018 17:00.
Plus d'informations sur https://foo.bar/baz"
);
$text->enableEncodingDetection();
$client->message()->send($text);
```

`$text->enableEncodingDetection();` should return `true` but currently it returns `false`.

We have tried with `new Unicode` and it works all the time.  
And of course I don't want to use that each time :)

But with the current version of library it doesn't take it into account.
This is why I take only your GSM alphabet in this P.R.

### DIFF
--- a/src/Message/EncodingDetector.php
+++ b/src/Message/EncodingDetector.php
@@ -12,18 +12,26 @@ class EncodingDetector {
 
     public function requiresUnicodeEncoding($content)
     {
+        $gsmCodePoints = array_map(
+            function ($char) {
+                $k = mb_convert_encoding($char, 'UTF-16LE', 'UTF-8');
+                $k1 = ord(substr($k, 0, 1));
+                $k2 = ord(substr($k, 1, 1));
 
-        $gsmCodePoints = [
-            0x0040, 0x00A3, 0x0024, 0x00A5, 0x00E8, 0x00E9, 0x00F9, 0x00EC, 0x00F2, 0x00E7, 0x000A, 0x00D8, 0x00F8, 0x000D, 0x00C5, 0x00E5, 0x0394,
-            0x005F, 0x03A6, 0x0393, 0x039B, 0x03A9, 0x03A0, 0x03A8, 0x03A3, 0x0398, 0x039E, 0x00A0, 0x000C, 0x005E, 0x007B, 0x007D, 0x005C, 0x005B,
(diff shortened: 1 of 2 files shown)


mheap:
Thanks for this! Based on the test you updated, it looks like `\n` and `\r` should be treated as unicode, whilst I'm not sure that's the case. @sammachin any ideas?

roukmoute:
Yep @mheap this is exactly what happens [in the form](https://developer.nexmo.com/messaging/sms/guides/concatenation-and-encoding#ConcatenationComponent).

If you just add a return carriage in the default text, you will see a big difference.

mheap:
@roukmoute I believe that the GSM 7bit alphabet does support CR LF so it shouldn’t need Unicode. This may be an issue with the form in the documentation. We're looking in to it

mheap:
Hey @roukmoute, thanks for your patience. CR and LF are part of the GSM7 character set, which means that there's a bug in Nexmo Developer's testing tool rather than this library.

The rest of your refactoring looks great though - would you remove CR LF from the unicode set so that we can merge your refactoring?

roukmoute:
I've just read that:

```
The ETSI GSM 03.38 specification shows an uppercase C-cedilla
glyph at 0x09. This may be the result of limited display
capabilities for handling characters with descenders. However, the
language coverage intent is clearly for the lowercase c-cedilla, as shown
in the mapping below. The mapping for uppercase C-cedilla is shown
in a commented line in the mapping table.
```

mheap:
I'd treat it as a lowercase `C-cedilla` as that's what the spec defines it as

```
0x09	0x00E7	#	LATIN SMALL LETTER C WITH CEDILLA
```

roukmoute:
@mheap up

mheap:
Thanks for working with us on this @roukmoute! I'll tag a release now

PR #142, opened 2018-10-27
Update ClientTest.php

mohitgauniyal19a:
Fix error codes


mheap:
Thanks for the PR. Those comments are about the API itself, not our tests, which means that this PR is actually breaking our tests

PR #146, opened 2018-10-30
Add support for hmac related request signing

mheap:


src/Client/Signature.php
```diff
@@ -51,11 +53,25 @@ public function __construct(array $params, $secret)
...
+                return md5($data);
+                break;
+            case 'md5':
+            case 'sha1':
+            case 'sha256':
+            case 'sha512':
+                return hash_hmac($signatureMethod, $data, $secret);
+                break;
+            default:
+                throw new Exception('Unknown signature algorithm: '.$signatureMethod);
```
judy2k: May I suggest an improvement to this?
`"'Unknown signature algorithm: '.$signatureMethod"`
:arrow_down:
`'Unknown signature algorithm: '.$signatureMethod.'. Expected: md5hash, md5, sha1, sha256, or sha512'`

lornajane:
@mheap @judy2k This works well on sending messages for me, I've added a little bit to the README so that people will know how to use it - have a quick read but I think we're ready to go with this

mheap:
I've confirmed with support that this is SMS only. I'm happy to merge if you are. Can you take care of a release @lornajane?

PR #152, opened 2018-12-14
Add prefix pricing endpoint

mheap:


mheap:
@lornajane README examples added. Also spotted a bug in my assumption that each prefix == 1 country, so I've updated the code to return a list

mheap:
@lornajane Good catch. The exception was thrown for the assumption that were would be a single instance as we don't usually return `null`. Now that it's a list of countries with that prefix we can return an empty array. Updated the PR

PR #154, opened 2018-12-17
Verify message signatures

### LINKED ISSUE #90: Is there support for verifying signed messages?
Its not clear from the docs how or if you should verify inbound message signatures. Is this possible? or something I need to implement myself?

lornajane:
The code was already in the codebase but needed a tiny tweak and some documentation so people can find it. Fixes #90

### DIFF
--- a/src/Client/Signature.php
+++ b/src/Client/Signature.php
@@ -107,7 +107,12 @@ class Signature
     /**
      * Check that a signature (or set of parameters) is valid.
      *
-     * @param array| string $signature
+     * First instantiate a Signature object: this will drop any supplied
+     * signature parameter and calculate the correct one. Then call this
+     * method and supply the signature that came in with the request.
+     *
+     * @param array| string $signature The incoming sig parameter to check 
+     *      (or all incoming params)
      * @return bool
      * @throws \InvalidArgumentException
      */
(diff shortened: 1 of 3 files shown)


src/Client/Signature.php
```diff
@@ -121,7 +126,7 @@ public function check($signature)
             throw new \InvalidArgumentException('signature must be string, or present in array or parameters');
         }
 
-        return $signature == $this->signed['sig'];
+        return strtolower($signature) == strtolower($this->signed['sig']);
```
mheap: Have we checked that this is required and doesn't increase the chance of collisions?

lornajane: The signature generated by our code has the characters in lower case. The signature that arrives with an incoming message has the alpha parts in upper case. The check therefore was always failing until I added this change. I could argue that we should instead make our signature generation code output upper case since that's what the Nexmo servers seem to do, which would you prefer?

mheap: I'd rather update our code to match the server I think. Do we know why ours is lowercase and the server is uppercase?

lornajane: No but I've checked all the signature types and made the PHP library match

PR #160, opened 2019-03-07
Add ability to check and update account settings

lornajane:
This was missing from the library so I added it


afolson:
Looks good to me 👌(I wasn't assigned as reviewer so I'm not merging :D)

Issue #147, opened 2018-11-02
Nexmo\Insights\Client doesnt handle error responses well


runawaycoin:
Hi,

The Nexmo\Insights\Client class doesnt seem to handle error responses well.

For example
Following code example from here:
https://developer.nexmo.com/number-insight/building-blocks/number-insight-basic

```
 $basic  = new \Nexmo\Client\Credentials\Basic(NEXMO_API_KEY, NEXMO_API_SECRET);
 $client = new \Nexmo\Client($basic);
 $insights = $client->insights()->basic('07999999999');
```

The above will fail - as the number is not in internalised format (it would also be great if the basic function accepts the $additionalParams argument like the generic makeRequest function does) - but php code errors with:
```
Notice: Undefined index: national_format_number
in vendor\nexmo\client\src\Insights\Client.php (line 27)
```
Which is:
```
    $basic = new Basic($insightsResults['national_format_number']);
```

The server API response does not contain the key 'national_format_number'.

Additionally if the Nexmo API response is not 200, the code fails with this:
```
Argument 1 passed to Nexmo\Insights\Client::getException() must be an instance of Nexmo\Insights\ResponseInterface, instance of GuzzleHttp\Psr7\Response given, called in vendor\nexmo\client\src\Insights\Client.php on line 96
```

Thanks very much - i could write up a PR with these implemented / corrected if you would prefer?

mheap:
@runawaycoin A PR would be very much appreciated!

TheOtherBrianJ:
> The server API response does not contain the key 'national_format_number'.

Yes, it appears the API response is:
```
{
    "status": 3,
    "status_message": "Invalid request :: Not valid number format detected [ 07999999999 ]"
}
```
What should the expected behavior of the client be when the `status` is a value other than `0`? Should it throw an exception like when a non-200 http status is received? 

It's also curious to me that this request returns a 200 http status instead of 400, given the error.

lornajane:
Thanks @TheOtherBrianJ there are some good observations here and we are always delighted to receive the pull request as well! I am looking forward to reviewing that one - I will also offer your comments about the expected status code to the API teams, but better client handling is something more immediately within our control :)

PR #162, opened 2019-03-11
Improve error handling in Insights\Client

TheOtherBrianJ:
This PR addresses one issue raised in #147:

> Additionally if the Nexmo API response is not 200, the code fails with this:
> `Argument 1 passed to Nexmo\Insights\Client::getException() must be an instance of Nexmo\Insights\ResponseInterface, instance of GuzzleHttp\Psr7\Response given, called in vendor\nexmo\client\src\Insights\Client.php on line 96`

The issue was a missing `use` statement.

I added a unit test to exercise the `getException` method and validate the fix.

I also moved the call to `$response->getBody()->getContents()` until _after_ the call to `getException`. Otherwise the attempt to `getContents` in `getException` wouldn't find anything as we'd be at the end of the stream.

Also did a check on the contents of the body before attempting to access `$body['error-code-label']`

### DIFF
--- a/src/Insights/Client.php
+++ b/src/Insights/Client.php
@@ -101,12 +101,13 @@ class Client implements ClientAwareInterface
     protected function getException(ResponseInterface $response)
     {
         $body = json_decode($response->getBody()->getContents(), true);
+        $body_error = isset($body['error-code-label']) ? $body['error-code-label'] : 'Unknown error';
         $status = $response->getStatusCode();
 
         if($status >= 400 AND $status < 500) {
-            $e = new Exception\Request($body['error-code-label'], $status);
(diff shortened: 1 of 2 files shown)


dragonmantank:
@TheOtherBrianJ Thanks for this! 

With our 2.0 release, we added some additional content to the exceptions that can be thrown. You should now be able to get the original response (`$e->getEntity()`), so you can handle direct errors or status codes. We also now should be throwing either a `Nexmo\Client\Exception\Request` or `Nexmo\Client\Exception\Server`, depending on which end caused the error, to make it easier for your code to handle.

Does this solve the use case for you?

If not, we can revisit this.

Issue #5, opened 2016-05-25
Currently Tied to the HTTPlug Guzzle Adaptor


tjlytle:
For ease of installation, the HTTPlug Guzzle 6 adapter is in `require`. You can certainly add another adapter and use that when creating the client, but that leaves unused code under `/vendor` and provides no good way to remove the Guzzle dependency if there's some version conflict (say another package using Guzzle 5).

Not having any adapter required would mean installation took two commands (with the first replaceable by another adaptor):

```
composer require php-http/guzzle6-adapter
composer require nexmo/client
```

Is that an issue? Could a composer hook be used to interactively prompt for an adaptor if one doesn't exist?

Nyholm:
Yes, If you require the virtual package [php-http/client-implementation](https://packagist.org/providers/php-http/client-implementation) then composer will fail and ask you to install a HTTPlug client. See http://docs.php-http.org/en/latest/httplug/library-developers.html

Nyholm:
You could also use http-php/discovery to remove the dependency to zendframework/zend-diactoros

tjlytle:
@Nyholm Yeah, understand the value of using `php/client-implementation`. Main concern is having to communicate that you need to both `require nexmo/client` and pick an adaptor. Shipping with one by default helps solve that, but certainly not a perfect solution. 

Haven't fully explored this, but do you know if there's any way to interrupt composer's error (when missing a client implementation), and dynamically add guzzle as a default?

I did play with `http-php/discovery`, but found puli kind of hard to work with / understand. I didn't take a lot of time with it though.

Nyholm:
Depending on php-http/client-implementation will interrupt composer if a client is missing, but you can't provide a default one. You could say something like: 

> to install this lib run `composer require nexmo/client php-http/guzzle6-adaper`

Have you seen the discovery lib after version 1.0? Puli is no longer a requirement. :)

tjlytle:
>  Puli is no longer a requirement. :)
I'll take another look.

Nyholm:
I wrote it myself. I'm pretty sure :)
You can use puli but it is no requirement.

mheap:
I had a chat with @tjlytle about this yesterday. We're planning on renaming `nexmo/client` to `nexmo/client-core` and have `nexmo/client` be an empty package that requires `nexmo/client-core` and `php-http/guzzle6-adapter`.

Any reason that this wouldn't work?

Nyholm:
[...]

PR #163, opened 2019-03-17
Upgrade dependencies

benja-M-1:
Hi there :)

I upgraded the version of the guzzle adapter because I needed it to create an API client in my project.

It would be even better to make this client to allow to use any client as proposed in https://github.com/Nexmo/nexmo-php/issues/5. But for now it works :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,13 @@
   "require": {
     "php": ">=5.6",
     "php-http/client-implementation": "^1.0",
-    "zendframework/zend-diactoros": "^1.3",
-    "php-http/guzzle6-adapter": "^1.0",
+    "zendframework/zend-diactoros": "^2.1",
+    "php-http/guzzle6-adapter": "^2.0",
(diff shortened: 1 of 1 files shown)


lornajane:
Thanks for sharing this with us - we know this is an area that needs some attention and a pull request and an explanation of why it's usefully really helps :)

benja-M-1:
@lornajane I don't understand if you expect something from me or not :)

Tests are failling because the guzzle6-adapter requires php >= 7.1. Do you want to support PHP < 7.1? In this case then a more complex PR must be sent.

benja-M-1:
@lornajane sorry to bother you but I prefer to strike while the iron is hot :) 
I submitted #164 that should be farther and is a better fix.

lornajane:
Thanks for the update. I am not too worried about the PHP 7.1 requirement - there are one or two other things on the roadmap that will require this soon anyway. I see there's a replacement PR so I'm closing this one and reviewing that one ....

PR #164, opened 2019-03-18
Make Client Guzzle agnostic

benja-M-1:
This PR replaces #163.

The point here is to decouple the Client from Guzzle to allow end users to use whatever http client implementation they want.  

Long story short, I needed to create an API for another tool relying on HTTPlug 2.0 but this library did not allow me to use it because it relies on guzzle-adapter 1.6.3. 

The build wont pass because nexmo/client build is configured to run on php 5.6 and 7.0. But Guzzle Adapters does not support PHP <7.1. Do you want me to remove support for those versions as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ To use the client library you'll need to have [created a Nexmo account][signup].
 To install the PHP client library to your project, we recommend using [Composer](https://getcomposer.org/).
 
 ```bash
-composer require nexmo/client
+composer require nexmo/client php-http/guzzle6-adapter
 ```
 
 > You don't need to clone this repository to use this library in your own projects. Use Composer to install it from Packagist.
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     "php-http/client-implementation": "^1.0",
     "zendframework/zend-diactoros": "^2.1",
     "php-http/guzzle6-adapter": "^2.0",
-    "lcobucci/jwt": "^3.2"
+    "lcobucci/jwt": "^3.2",
+    "php-http/discovery": "^1.6"
   },
   "require-dev": {
     "phpunit/phpunit": "^5.7",
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,6 @@
     "php": ">=5.6",
     "php-http/client-implementation": "^1.0",
     "zendframework/zend-diactoros": "^2.1",
-    "php-http/guzzle6-adapter": "^2.0",
     "lcobucci/jwt": "^3.2",
     "php-http/discovery": "^1.6"
   },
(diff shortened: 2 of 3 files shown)


lornajane:
Thanks again for another PR! It's good to get input from others who also use our library :)

It would be good to bring some of these dependencies up to date although we don't want to set high compatibility levels for no reason. It is also fine to require PHP 7.1 - we will be bumping our minimum requirement soon anyway since nobody should be starting new projects on older versions and there's a feature I want to add that will need it.

That said, I am afraid we can't accept this pull request. There are lots of tutorials that use this library and users may not look to the latest `README` for installation instructions so requiring two libraries to be installed via Composer rather than one will trip up more people than it solves problems for. The approach outlined in #5 is still my preferred approach (I know that looks idle, we're still actively talking about it I promise!) so that a simple install works for most people and those who need something else can swap out the adapter.

Before I close this pull request, is there an approach that would help achieve what you need right now, before we get the "proper" changes made?

benja-M-1:
> That said, I am afraid we can't accept this pull request. There are lots of tutorials that use this library and users may not look to the latest README for installation instructions so requiring two libraries to be installed via Composer rather than one will trip up more people than it solves problems for. The approach outlined in #5 is still my preferred approach (I know that looks idle, we're still actively talking about it I promise!) so that a simple install works for most people and those who need something else can swap out the adapter.

I understand that this change requires that the documentation must be updated. Is it a lot of work? Can I help on that?

Nevertheless, as I added the "suggest" keyword on the composer json, when installing the client, composer will tell the end user that he should install guzzle-adapter too. 
Furthermore, [the Psr18ClientDiscovery.php](https://github.com/php-http/discovery/blob/master/src/Psr18ClientDiscovery.php#L27) throw an exception when it cannot find any client. 
So the code is self documented and thus I am not sure updating the docs everywhere is needed.

> Before I close this pull request, is there an approach that would help achieve what you need right now, before we get the "proper" changes made?

If you don't want to pursue the discussion to merge this, then my previous PR is fine to fix the issue I raised :)

benja-M-1:
Hi @lornajane how can we move on on this subject?

benja-M-1:
Hi @lornajane any news?

lornajane:
Hey @benja-M-1 , we have a few things in flight with the PHP library just now - I am leaving this open so we don't forget it but I think maybe merging #163 will make more sense once we've dealt with the other changes that are coming up. Thanks for your input on this!

lornajane:
Yes please! I'm totally in favour of this change, for most people the current setup is fine but by moving the packages around we allow more flexibility for people that need it! I'd rather do it this way than change the installation instructions in case people are following older existing instructions.

lornajane:
OK. I'm definitely closing this but we've done a refactor to solve the problem by renaming the packages as suggested in the discussion thread here. You can see what's going on if you look at #182  ... any feedback anyone has on this approach is very gratefully received. Also: thanks again for helping us understand the changes we needed here, and helping to "encourage" us to work on them :) We hope you love these changes when we ship them!

PR #169, opened 2019-04-23
Add custom guzzle client example to docs

lornajane:
I keep looking up an old issue to see how to do this, so I thought I should add it to the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,10 +599,10 @@ curl.cainfo = "C:\php\extras\ssl\cacert.pem"
 
 We allow use of any HTTPlug adapter, so you can create a client with alternative configuration if you need it, for example to take account of a local proxy, or deal with something else specific to your setup.
 
-Here's an example, this one disables the verification step, which is not recommended but does make a good example!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -592,6 +595,17 @@ curl.cainfo = "/etc/pki/tls/cacert.pem"
...
 ```
 
+### Pass custom Guzzle client
+
+We allow use of any HTTPlug adapter, so you can create a client with alternative configuration if you need it, for example to take account of a local proxy, or deal with something else specific to your setup.
+
+Here's an example, this one disables the verification step, which is not recommended but does make a good example!
+
+```
+$adapter_client = new Http\Adapter\Guzzle6\Client(new GuzzleHttp\Client(['verify' => false]));
```
mheap: Is there any other example we could add here? This one makes me nervous

lornajane: Fair enough, I just picked the one from the issue because we knew it worked. Replaced with a sensible timeout instead

Issue #171, opened 2019-04-25
Unable to use with Laravel Passport (and likely other packages)


bstump:
This package is used in Laravel's "laravel/nexmo-notification-channel package". Due to the Zend requirement in the composer file of the nexmo/nexmo-php package, an error is being thrown when I try to install this package:

"zendframework/zend-diactoros": "^1.3"

It looks like other packages were having this issue as well and were able to resolve it by allowing multiple versions of Zend Diactoros:

"zendframework/zend-diactoros": "^1.3|~2.0"

I have not pulled and tested myself, but perhaps you could take a look so I can use this with my other packages?

Please let me know. Thanks!

gcshri:
Seeing this error:

````
 Problem 1
    - Conclusion: don't install laravel/nexmo-notification-channel v2.0.0
    - Conclusion: remove zendframework/zend-diactoros 2.1.1
    - Installation request for laravel/nexmo-notification-channel ^2.0 -> satisfiable by laravel/nexmo-notification-channel[2.0.x-dev, v2.0.0].
    - Conclusion: don't install zendframework/zend-diactoros 2.1.1
````

**Comment by arubacao** [comment]
@gcshri and @bstump : I had similar problems and fixed them in this branch: https://github.com/arubacao/nexmo-php/tree/feature/php7-update

This does the trick:

```
{
    "name": "arubacao/test",
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/arubacao/nexmo-php"
        }
    ],
    "require": {
      "nexmo/client": "dev-feature/php7-update"
    }
}
```

josegonzalez:
We're also seeing this issue in cakephp/cakephp#13229.

lornajane:
Thanks @josegonzalez, we'll keep you posted

lornajane:
Should be fixed with the 1.8.1 release so I'll close the issue but please shout if you see anything else we should know about. Thanks all!

Issue #167, opened 2019-04-15
Security warning about composer.json


repat:
We're using [roave/security-advisories]( https://packagist.org/packages/roave/security-advisories) to not accidentally install PHP packages with known security issues. When installing `nexmo/laravel-notification` I got a warning and couldn't install. Following this further down the rabbit hole I got:

> nexmo/laravel-notification => nexmo/laravel => nexmo/client => zendframework/zend-diactoros: ^1.3

Seems like it's fixed in `zendframework/zend-diactoros: ^1.8.4`, at least according to roave, I didn't do that much research. Technically it should actually install 1.8.6 but doesn't if you already installed version 2.x. Maybe this could be prevented by putting `^1.8.5` in this `composer.json`?

lornajane:
Thanks for flagging this for us! I'll take a look

lornajane:
Should be fixed in 1.8.1. Thanks for pointing this out, we really appreciate it :)

PR #175, opened 2019-05-07
Bump ZF diactoros version after security alert

lornajane:
It doesn't seem to require a significant bump in other requirements, and still supports PHP 5.6.

We've had this mentioned a few times: #167 saw it coming from the Laravel component, as did #171. It's mentioned in #151 (although there are other packages there).


mheap:
Should we also allow `~2.0` so that #171 is resolved?

lornajane:
Yes, added to the patch.

PR #181, opened 2019-05-21
Update composer.json

serderovsh:


lornajane:
Thanks so much for offering us this pull request. We're actually refactoring the library to remove the requirement for the guzzle adapter - you can see what we're working on in #182 is you're interested. I've also increased the version of the adapter that we'll install if users want one.

Since we're already working on this, I'll close this pull request but thank you so much for sending it over and taking an interest in our library, it's greatly appreciated!

PR #179, opened 2019-05-17
[DO NOT MERGE] Feature/php7 update

lornajane:
This branch from @arubacao contains a bunch of really valuable fixes/updates. We can't accept it exactly as it is and I'm not sure we want everything that's here, but I found the diff interesting so I'm opening it as a PR to make it easy to diff and discuss the changes. 

@arubacao did you use a particular tidy/formatting tool for this? I am aware of a few but whatever this was it gets good results at weeding out unnecessary `use` statements etc

PR #182, opened 2019-06-04
Separate Core and Wrapper Composer Packages

### LINKED ISSUE #5: Currently Tied to the HTTPlug Guzzle Adaptor

lornajane:
Currently, nexmo/client includes a dependency on `php-http/guzzle6-adapter` but in fact the Nexmo functionality can be used with any HTTP client that satisfies `php-http/client-implementation`. This pull request aims to improve this situation by:

* renaming the current PHP package to `nexmo/client-core` and removing the guzzle adapter dependency
* creating another package `nexmo/client` which requires `nexmo/client-core` and `php-http/guzzle6-adapter` - the new package is here https://github.com/Nexmo/nexmo-php-complete

For most users: existing instructions `composer install nexmo/client` will still be correct.

For users with a dependency problem or wishing to use a different HTTP client: Those users can install `nexmo/client-core` which will fail unless they also install a compatible HTTP client. This is documented in the README.

This fixes #5 and also aims to improve the situation that #163 and #164 were intended to fix. The version of Guzzle adapter has also been bumped since that has been causing some dependency problems (see #179 also).


lornajane:
The builds failed because the tests rely on the Guzzle adapter. Does it make sense to add this as a dev dependency? Or maybe there is a more minimal HTTP client we could add instead.

PR #183, opened 2019-06-06
fix: update guzzle version to 2.x.x

atymic:
This is just a reminder PR to myself to fix the dependency in our app once this is fixed in the main repo.


atymic:
Oops, this want meant to be into my fork :see_no_evil: 
Sorry.

Issue #168, opened 2019-04-16
PHP Unit Mock Objects is deprecated


lornajane:
The `composer install` command shows this message `Package phpunit/phpunit-mock-objects is abandoned, you should avoid using it. No replacement was suggested.` We should look at our use of this package (and our version of PHPUnit since we're a bit behind) to resolve this message.

dragonmantank:
PHPUnit was updated and this message has since gone away. Closing.

PR #185, opened 2019-06-14
Bump to newer PHPUnit

lornajane:
We're wildly out of date with our PHPUnit versions, which isn't an emergency but isn't great practice either. I've started by bumping the version from 5.7 to 6 (I have 6.5 locally but I don't think it's critical) and fixing all the tests that needed attention. This makes us a mere 3 years behind in PHPUnit versions :)


lornajane:
Build failing because the travis file still has PHP 5 in it

test/Redact/ClientTest.php
```diff
@@ -85,11 +85,10 @@ public function testOptionsDoNotOverwriteParams()
 
     /**
      * @dataProvider exceptionsProvider
+     * @expectedException Nexmo\Client\Exception\Request
```
mheap: We lose a test by moving this to an annotation. `setExpectedException` is deprecated but we can use `$this->expectException($expectedException);` and `$this->expectExceptionMessage ($expectedMessage);` instead

lornajane: Thanks! I added this

lornajane:
We're stopping here: PHPUnit v8 is out but requires PHP 7.2, we're minimum version PHP 7.1 so let's stick on PHPUnit 7.4+ but <8.0 for now.

lornajane:
Also fixes #168 as a happy side effect

PR #186, opened 2019-06-20
[WIP] Error handling improvements and response data in Exceptions

### LINKED ISSUE #147: Nexmo\Insights\Client doesnt handle error responses well

lornajane:
This PR has breaking changes, and must be released into a v2 library version.

I added error handling for the Number Insights API - this did not have tests for errors but I have now added some. This change fixes #147 .

For other areas, I have added the response with a `setEntity()` call on the Exception object. This is available on `Exception\Request` and `Exception\Server` but not on the generic `Exception\Exception` object so there are some places where the exception type also changed where I felt the response body would be particularly useful. It's a string so needs rewinding to be usable.

The aim is to make the response data more readily available to users, there's more we could do but I think this is the last work blocking a v2 release.


davidbankes:
I just ran into a less-than-helpful exception:

unexpected response from API on line 54 of /[...]/vendor/nexmo/client/src/Message/Client.php
#0 /home/bitemeals/releases/live/vendor/laravel/framework/src/Illuminate/Notifications/Channels/NexmoSmsChannel.php(61): Nexmo\Message\Client->send(Object(Nexmo\Message\Message))

On line 54 of that file, it checks to see if the response data has a key and if not, throws an exception with no details about what the response was. This makes it hard to understand what went wrong.

Seems like this would be a nice fix to add to this PR.

lornajane:
Thanks @davidbankes this was an excellent time to point this out to us indeed :) I've added some additional error handling so that you should be able to get the response data by calling the `getEntity()` method on the Exception object in this situation,

PR #188, opened 2019-06-27
Remove beta indications, we're definitely not in Beta

lornajane:
I noticed (on packagist, where the formatting is different) that the library says we're still in beta. We're definitely stable now so I updated that, and I tidied up a few other things while I was here:

* Remove license section, we have the badge and also indicate our license in the standard GitHub way with a LICENSE file which I think users expect now
* Remove API coverage section, we've got all the coverage now

### DIFF
--- a/README.md
+++ b/README.md
@@ -695,3 +695,13 @@ Contributing
 
 This library is actively developed and we love to hear from you! Please feel free to [create an issue][issues] or [open a pull request][pulls] with your questions, comments, suggestions and feedback.
 
+[signup]: https://dashboard.nexmo.com/sign-up?utm_source=DEV_REL&utm_medium=github&utm_campaign=php-client-library
+[doc_sms]: https://developer.nexmo.com/messaging/sms/overview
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -690,74 +689,9 @@ $adapter_client = new Http\Adapter\Guzzle6\Client(new GuzzleHttp\Client(['timeou
...
-
-To contribute to the library, docs, or examples, [create an issue][issues] or [a pull request][pulls]. Please only raise issues
-about features marked as working in the [API coverage](#API-Coverage) as the rest of the code is being updated.
-
-License
--------
-
-This library is released under the [MIT License][license]
-
-[signup]: https://dashboard.nexmo.com/sign-up?utm_source=DEV_REL&utm_medium=github&utm_campaign=php-client-library
```
dragonmantank: I think this (and a good chunk of the rest) of these need to stay so that we can keep the link references.

lornajane: Oh good catch!

PR #199, opened 2019-09-19
Update composer.json

if4lcon:
Just moved guzzle adapter to require this is important if you are using

`composer install --no-dev`

all projects in production does this

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,14 +34,14 @@
     "php-http/client-implementation": "^1.0",
     "zendframework/zend-diactoros": "^1.8.4 || ^2.0",
     "lcobucci/jwt": "^3.2",
-    "ocramius/package-versions": "^1.4"
+    "ocramius/package-versions": "^1.4",
+    "php-http/guzzle6-adapter": "^1.0"
   },
   "require-dev": {
     "phpunit/phpunit": "^7.4",
     "php-http/mock-client": "^0.3.0",
     "estahn/phpunit-json-assertions": "^3.0.0",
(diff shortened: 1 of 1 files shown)


dragonmantank:
Hello!

Moving Guzzle to being a dev dependency was an intentional move. For a full history of "why", please check out https://github.com/Nexmo/nexmo-php/issues/5. The tl;dr is that the way we packaged it before caused conflicts with other libraries also supplying Guzzles, leading to dependency resolution issues.

That being said, you should be able to continue to use `nexmo/client` with no changes. That has moved over to https://github.com/nexmo/nexmo-php-complete, and is still packaged with Guzzle like before. Production installs should continue to still work after upgrading to 2.0, as Packagist should switch you to the new repo.

This Github repository is now `nexmo/client-core` in Packagist, so should not conflict with anything in production. You should only use this package if you want to supply your own HTTPlug-compatible client for our SDK to use. 

If you are having an issue, either with dependency resolution or upgrading from 1.x to 2.x, please open an issue. 

Thank you!

if4lcon:
I'm using nexmo-laravel and it works on dev but not on production I couldn't make it work

Class 'Http\Adapter\Guzzle6\Client' not found

solved it by adding guzzle adapter to dependency

dragonmantank:
@if4lcon Can you open an issue over at https://github.com/nexmo/nexmo-laravel and let me know what version of `nexmo/laravel` you are using? If possible, a copy of your `composer.lock` would help immensely in this case.

PR #207, opened 2020-01-02
Replace zend/diactoros with laminas/diactoros

### LINKED ISSUE #206: Replace zendframework/zend-diactoros with laminas/laminas-diactoros
The `zendframework/zend-diactoros` packages has been marked as abandoned as of the end of 2019. There is a community-supported open source continuation in `laminas/laminas-diactoros`. Composer is now showing warnings about this on `composer install`, so we should update to depend on the Laminas package rather than the Zend one.

The new package is under a different namespace so will require code updates, but means there won't be any conflicts if projects otherwise depend on the Zend package.

jaikdean:
This replaces the abandoned `zend/diactoros` package with the new community-supported `laminas/diactoros`. As of now, this is ostensibly the same package under a new namespace, so the change is just to `composer.json` and the relevant FQCNs.

Fixes #206


mangelsnc:
There are plans to merge this PR soon?

dragonmantank:
There are! Travel has pushed back some of my work, but this should all go out in the next release.

Let me know if it's causing any issues, but other than the notification (which is just that, a notification) that happens during `composer install` everything should continue to work like normal.

dragonmantank:
I've done the swap in 2b55506b7190e46647044f3cdbe21de31676b08f, which will be landing in v3.0.0.

I'll leave this PR open until that release hits, just for tracking.

dragonmantank:
Sorry this took so long, this is now part of the 2.2.0 release. I didn't realize this contained a bridge to handle the namespace change as well, otherwise I would have merged it a long time ago :(

PR #213, opened 2020-02-03
nexmo-php / examples /send.php

elsantito:


lornajane:
** @elsantito  Please immediately change your account credentials since they are visible in this pull request!**

This change makes the example _less_ useful so I'm afraid we can't accept the PR this time.

PR #218, opened 2020-03-26
Add stream support

assertchris:
Things to consider:
- [x] how do we deal with `stream` being called, when the option wasn't set in the first place?
- [x] how do we test this jank?
- [ ] who is the best Next dev advocate?
- [ ] when will corvid be over?
- [x] ~~possibly add adapters for other clients, add those clients to suggests~~ (not needed)


assertchris:
@dragonmantank I think this is ready for review.

Thanks to @djekl and @azjezz for the pair programming.

kristyna123:
U

dragonmantank:
This is getting merged into v3.0.0-alpha branch, since there was a rename it closed the PR. But this is going in over there!

PR #220, opened 2020-04-10
refactor Account/Price.php - use look up table and change if conditio…

ualiyou:
…n syntax

### DIFF
--- a/src/Account/Price.php
+++ b/src/Account/Price.php
@@ -20,28 +20,23 @@ abstract class Price implements EntityInterface, JsonSerializableInterface, Json
 
     protected $data = [];
 
-    public function getCountryCode()
-    {
+    public function getCountryCode() {
         return $this['country_code'];
     }
 
-    public function getCountryDisplayName()
-    {
+    public function getCountryDisplayName() {
         return $this['country_display_name'];
     }
 
-    public function getCountryName()
-    {
+    public function getCountryName() {
         return $this['country_name'];
     }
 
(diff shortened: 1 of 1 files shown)


src/Account/Price.php
```diff
@@ -69,76 +61,71 @@ public function getPriceForNetwork($networkCode)
...
-
-                if (isset($n['network'])) {
-                    $n['networkName'] = $n['network'];
-                    unset($n['network']);
-                }
-
-                $network = new Network($n['networkCode'], $n['networkName']);
-                $network->jsonUnserialize($n);
-                $networks[$network->getCode()] = $network;
+        if (!isset($json['networks'])) {
```
dragonmantank, line 89: While the short-circuit return is nice, functionally it won't help much. We're replacing one `if` evaluation with another, so I don't see much gain for doing this. We're also losing the fact that there are empty networks, which the previous version would have recorded.

ualiyou: Yes it's my phpstorm config for reformat codes and i change config now,
Can I help refactor code in the process?

dragonmantank: Yes, helping refactor the code is always appreciated!

We just want to make sure that our code follows a similar process, and my notes above were more how we can refactor this in a more complete way, but any help from the community is great.

dragonmantank:
Closing due to inactivity. If you'd like to work on this feel free to push an update with the changes requested above and and re-open the PR, we'll see what we can do!

Issue #227, opened 2020-06-12
Allow direct access to incoming MSISDN instead of just through `from`


dragonmantank:
When creating an inbound SMS object, we hide the `msisdn` parameter behind `from`, since they are the same thing. We should add support for getting the `msisdn` since that is the actual field name, but it will be the same value as `from` currently returns.

```php
$inbound = \Nexmo\Message\InboundMessage::createFromGlobals();
error_log($inbound['from']); // Current
error_log($inbound->getMsisdn()); // Proposed, as we're dropping array syntax for most objects
```

dragonmantank:
This is now available in the new `\Nexmo\SMS\Webhook\InboundSMS` class as part of 2.2.0. Even if the old `messages()` client is used to send the SMS, the new `sms()` client can handle the incoming data on the webhook.

```php
$inbound = \Nexmo\SMS\Webhook\Factory::createFromGlobals();
error_log($inbound->getMsisdn());
error_log($inbound->getFrom());
```

Issue #221, opened 2020-04-15
Sending an SMS hides the fact multiple messages were sent


dragonmantank:
From discussion on Discord with a user:

> I am working with SMS Message price, now SMS can consist of more than 1 message parts, right? concatenation and stuff. Nexmo charges for each SMS part.
>
> If we look at this getPrice method in nexmo-php , does it return the cost of Message as a whole? or just one part (last one)?

Digging into this more, the `Nexmo\Message\Message` object assumes a single message is returned by the API. In the case where multiple messages are returned, such as a long message that is split in two, it only returns data for the last chunk. The messages send properly, the SDK just doesn't report back data for the _other_ chunk.

The SDK should not lose the context of multiple messages like that, but this will be a breaking change to fix. 

Short-term fix is to use the raw request data:

```php
$sms = $nexmo->message->send(...);
$response = $sms->getResponseData();
$total = 0.0;
foreach ($response['messages'] as $message) {
    $total += (float)$message['message-price'];
}
```

awebartisan:
Maybe we can consider `BCMath` to perform the addition and not loose precision? 

- This extension does not come pre-packaged with PHP but can be installed easily.
- The Windows version of PHP has built-in support for this extension.

dragonmantank:
This is partially fixed in v2.2.0 in the new `sms()` client (which isn't available in Laravel yet, which is where the original problem was discovered).

The above suggestion is basically the same as the current solution, just the returned data from the API is more apparent that it is iterable.

```php
$response = $client->sms()->send(new SMS('15551111234', '15552227894', 'Test message'));
$total = 0.0;
foreach ($response as $message) {
    $total += (float) $message->getMessagePrice();
}
```

As for using the BC Math extension, the end user is free to implement this to keep float precision. I did not add a way to get the total cost because of this, but the new interface is much more up-front that the result from sending an SMS is an iterable object they can pull individual message details out of. 

I'm closing this, but leaving #219 open until the Laravel package is updated to support the newer `sms()` client, which will be a backwards-compatibility-breaking change.

Issue #219, opened 2020-04-09
Stream not rewinding after receiving Nexmo Message object in Laravel Nexmo Channel


awebartisan:
I am sending SMS using Nexmo and **laravel/nexmo-notification-channel.** I want to use the **response**  that gets back from Nexmo. 
Fortunately, Laravel provides an event `NotificationSent` and a public property on it called `response` which contains the object of Nexmo\Message\Message and I can see all the necessary data by dumping it.
Now, I want to get the **status** of message. So I go ahead and call the available method.

`$nexmoMessage->getStatus()`

But every messaage is throwing following exception:
```ErrorException: fseek() expects parameter 1 to be resource, int given in /Users/zubairmohsin/sites/sitename/vendor/guzzlehttp/psr7/src/Stream.php:204```

### Stacktrace
```

#1 /Users/zubairmohsin/sites/sitename/vendor/guzzlehttp/psr7/src/Stream.php(204): fseek(0, 0, 0)
--
#2 /Users/zubairmohsin/sites/sitename/vendor/guzzlehttp/psr7/src/Stream.php(190): GuzzleHttp\Psr7\Stream->seek(0)
#3 /Users/zubairmohsin/sites/cartrecovery/vendor/nexmo/client-core/src/Entity/JsonResponseTrait.php(29): GuzzleHttp\Psr7\Stream->rewind()
#4 /Users/zubairmohsin/sites/sitename/app/Domain/SMS/Listeners/MessageSent.php(36): Nexmo\Message\Message->getResponseData()
#5 [internal function]: App\Domain\SMS\Listeners\MessageSent->handle(Object(Illuminate\Notifications\Events\NotificationSent))
```

**Comment by dragonmantank** [comment]
Just tested with this sample code with Laravel 7 and PHP 7.4.4:

```php
Route::get('/nexmo', function () {
    Notification::route('nexmo', '55555555555')
        ->notify(new TestNexmoNotification())
    ;
});
```

```php
class TestNexmoNotification extends Notification
{
    use Queueable;

    /**
     * Get the notification's delivery channels.
     *
     * @param  mixed  $notifiable
     * @return array
     */
    public function via($notifiable)
    {
        return ['nexmo'];
    }

    /**
     * Get the mail representation of the notification.
     *
     * @param  mixed  $notifiable
     * @return \Illuminate\Notifications\Messages\MailMessage
     */
    public function toNexmo($notifiable)
    {
        return (new NexmoMessage)
[...]

Issue #177, opened 2019-05-09
Add throttling support?


mheap:
Should we add support for throttling endpoints that have rate limits? https://github.com/hamburgscleanest/guzzle-advanced-throttle looks like a reasonable solution

arubacao:
IMO if you allow custom clients (e.g. #164) I don't feel like this is necessary for this package. Maybe a tutorial on how to create a custom client that is capable of throttling is sufficient. 
However, in the same context, maybe revise this part of the code: https://github.com/Nexmo/nexmo-php/blob/2f79f67f24225ea627ee14578e98c96276cdd4c5/src/Message/Client.php#L63-L69
Rather fail the request than have a never ending retry loop that cannot really be controlled anymore (and potentially cost money(?)).

lornajane:
I think that we probably don't want to add throttling or retries at the library level, but I think we could improve the error handling which would make it easier for users to understand if something didn't work and take whatever action makes sense in their context? The point about getting stuck in the endless retry loop is a good one!

dragonmantank:
As of v2.2.0, there is now a specific `Nexmo\Client\Exception\ThrottleException` that is thrown.

In the few places it made sense or we already had support, the Nexmo client now watches for this specifically and handles it. This is specifically in the SMS and Numbers where it is incredibly easy to hit the throttle limit without knowing.

PR #230, opened 2020-07-03
v2.2.0 Release

dragonmantank:
Flagged as `dont-merge` while performing RC testing, but this should be the final release code. Code snippets have been prepared to support all the new v2.2.0 changes. 

This release focuses on deprecation of dead and old code, and preps many internal changes in regards to v3.0.0. Where possible upcoming v3.0.0 changes were backported where backward-compatibility could be maintained. This code is the same as earlier v2.2.0 branches and PRs, but cleaned up to better support the RC release.

### Testing

If you would like to test this with an existing application, remove `nexmo/client` and replace it with this version of core:

```bash
composer remove nexmo/client
composer require "php-http/guzzle6-adapter":"^2.0" "nexmo/client-core":"v2.2.0.x-dev@dev"
```

No other code changes should be required. If you get any errors or the application begins to misbehave, that indicates a bug, please let me know.

### Added

* New Voice and SMS interfaces, accessible through `$client->voice()` and `$client->sms()`, respectively
* Added user deprecation warnings, which can be turned on and off via the `Nexmo\Client` "show_deprecations" config option. This can help devs update in preparation in v3.0.0, and will be utilized in the future as things are deprecated.
* Many objects now have a `toArray()` serialization method, to discourage direct use of `jsonSerialize()`
* Many objects now support a `fromArray()` hydration method
* Better incoming webhook support for SMS and Voice events
* NCCO builder for Voice

### Changed

* API handling code has been conglomerated in `Nexmo\Client\APIResource` and `Nexmo\Entity\IterableAPICollection`
* All APIs ported over to the new API handling layer
* Internal Service Locator `Nexmo\Client\Factory\MapFactory` is now PSR-11 compliant, and can use factories
* Most Verify methods in the client now prefer string Request IDs
* Verify now prefers `Nexmo\Verify\Request` for starting new Verification requests

### Deprecated
For a detailed list of things that may impact an application, enable the `show_deprecations` `Nexmo\Client` option to see deprecation notices for specific code flows.

* Most forms of array access are now deprecated
* Using service layers like `$client->messages($filter)` to search has been deprecated in favor of bespoke search methods
* Requests/Responses on exceptions and entities are deprecated, and now exist in the service layer or collection
* Most methods that took raw arrays now prefer objects
* `Nexmo\Verify\Verification` objects full functionality has been deprecated, and will be used only as a value object for search requests in the future
* `Nexmo\Conversations` and `Nexmo\User` have been deprecated and will be removed in the future as the feature is in Beta status and has diverged from this implementation
* `Nexmo\Voice\Call` and `Nexmo\Voice\Message` have been deprecated and will be removed in the future as the TTS API is deprecated
* SMS searching has been deprecated and will be removed in a future update

### Removed
* No features or classes have been removed in this update, it is functionally compatible with v2.1.0 other than deprecation notices and new features.

### Fixed
* No direct bugs have been fixed as this release was designed to be as compatible with v2.1.0 as possible, however:
  * #177 should be better handled by a centralized `Nexmo\Client\Exception\ThrottleException` and has been implemented in SMS and the Numbers API
  * #219 is implicitly fixed in `Nexmo\SMS\Client::send()` as it now returns a fully hydrated collection object as a response, however this needs to be updated in Laravel itself via an update to `nexmo/laravel` and `laravel/nexmo-notification-channel `
  * #221 is implicitly fixed in `Nexmo\SMS\Client::send()` as it now returns a fully hydrated collection object that is much more up-front it is not a single object
  * #227 is implicitly fixed in `Nexmo\SMS\Webhook\InboundSMS`
  
### Security

* There were no known security vulnerabilities reported


src/Client.php
```diff
@@ -108,18 +113,42 @@ public function __construct(CredentialsInterface $credentials, $options = array(
...
+            'sms' => \Nexmo\SMS\ClientFactory::class,
+            'verify' => \Nexmo\Verify\ClientFactory::class,
+            'voice' => \Nexmo\Voice\ClientFactory::class,
+
+            // Additional utility classes
+            APIResource::class => APIResource::class,
         ], $this));
+
+        // Disable throwing E_USER_DEPRECATED notices by default, the user can turn it on during development
+        if (array_key_exists('squelch', $this->options) && $this->options['squelch']) {
```
mheap, line 138: Can we think of a better option name?

mheap:
How does the NCCO builder compare to what we have in other languages? Should we have an RFC for builders?

src/Client.php
dragonmantank: I'm open to ideas. I went through a few before settling on squelch, for [the electronic reason](https://www.lexico.com/en/definition/squelch), not the sound effect reason.

dragonmantank:
> How does the NCCO builder compare to what we have in other languages? Should we have an RFC for builders?

Stop looking at my wish lists! 

With this being the third NCCO builder, it did cross my mind.  This is close to what we have in Java and .NET. Actions are added to a main builder, and the result can be output as the appropriate JSON entity. 

## PHP
```php
$ncco = new \Nexmo\Voice\NCCO\NCCO();
$ncco
    ->addAction(
        new \Nexmo\Voice\NCCO\Action\Talk('Welcome to the amazing Nexmo conference call')
    )
    ->addAction(
        new \Nexmo\Voice\NCCO\Action\Conversation('amazing-conference-call')
    )
;

header('Content-Type: application/json');
$json = json_encode($ncco);
echo($json);
```

### .NET
```csharp
var talkAction = new TalkAction() { Text = "Please wait while we connect you to the conference" };
var conversationAction = new ConversationAction() { Name = CONF_NAME };
var ncco = new Ncco(talkAction, conversationAction);
return ncco.ToString();
```

### Java
```java
TalkAction intro = TalkAction.builder("Please wait while we connect you to the conference.").build();
ConversationAction conversation = ConversationAction.builder(CONF_NAME).build();
res.type("application/json");
return new Ncco(intro, conversation).toJson();
```

src/Client.php
mheap: `show_deprecations`? Does what it says on the tin

dragonmantank: Updated to `show_deprecations`, which defaults to `false` to preserve existing behavior.

dragonmantank:
Just noting here from the test I ran as part of "The Late Night Build" hackathon:

- Only issues presented with the code snippets was with SMS sending, `BRAND_NAME` throws US people off because it isn't clear you have to use a telephone number
- Incoming webhook parser for SMS only encountered a problem due to ngrok configuration
- Actual voice calls and SMS portions worked fine

examples/delete-application.php
```diff
@@ -4,7 +4,17 @@
...
 
-$a = $client->applications()->get(APPLICATION_ID);
+try {
+    $isDeleted = $client->applications()->delete(MESSAGES_APPLICATION_ID);
 
-$client->applications()->delete($a);
+    if ($isDeleted) {
+        echo "Deleted application " . MESSAGES_APPLICATION_ID . PHP_EOL;
+    } else {
+        echo "Could not delete application " . MESSAGES_APPLICATION_ID . PHP_EOL;
```
slorello89, line 13: Doesn't seem like this is reachable- delete returns a 204 so how would we know if it's deleted or not? should error out in a 401/404

dragonmantank: This is true, I'll need to update the snippet as well. The `try/catch` is the only way to know this doesn't really work.

examples/make-call-ncco.php
```diff
@@ -2,23 +2,21 @@
...
-  ->setNcco([
-        [
-            'action' => 'talk',
-            'text' => 'This is a text to speech call from Nexmo'
-        ]
-      ]);
-
-$response = $client->calls()->create($call);
+$outboundCall = new \Nexmo\Voice\OutboundCall(
+    new \Nexmo\Voice\Endpoint\Phone(TO_NUMBER),
```
slorello89, line 11: I notice there's a lot of fully qualified namespacing - is that considered common practice in PHP? Would it maybe make it cleaner to just pull in the namespace or is that a big negative on performance?

dragonmantank: Holdover from code snippets. Normally code would be dragged in with a `use` statement at the top to help alias these out.

slorello89: In that case would it be prudent to drag in the namespace so that the snippet looks much cleaner? I'll bring your attention to the new import dependencies section of the code snippets.

src/Account/Balance.php
```diff
@@ -68,4 +84,28 @@ public function offsetUnset($offset)
...
+    {
+        $this->data = [
+            'balance' => $data['value'],
+            'auto_reload' => $data['autoReload']
+        ];
+    }
+
+    public function toArray(): array
+    {
+        return $this->data;
```
slorello89, line 98: Am I reading this rightt? looks like `data` is a dictionary?

src/Account/Balance.php
```diff
@@ -68,4 +84,28 @@ public function offsetUnset($offset)
...
+    }
+
+    public function __get($key)
+    {
+        if ($key === 'data') {
+            trigger_error(
```
slorello89, line 104: This looks like a BC - does this belong in 2.2.0?

src/Account/Balance.php
dragonmantank: Yeah. PHP arrays pull double-duty as both traditional arrays and dictionaries. So this object stores everything as an internal dictionary/array, so we just return it.

src/Account/Balance.php
dragonmantank: No, this actually preserves behavior because of a visibility change.

The original usage allowed users to directly access `$balance->data` for things as `data` had Public visibility. That is now Protected visibility, so this uses `__get()` as a catch-all getter. If they request "data" as they used to previously, we'll return it but throw the deprecation error.

examples/make-call-ncco.php
dragonmantank: Yup, I'll fix these here, and then update the code snippets since NDP has been updated to handle imports.

src/Application/Application.php
```diff
@@ -201,9 +222,8 @@ public function jsonUnserialize(array $json)
         }
     }
 
-    public function jsonSerialize()
```
slorello89, line 204: looks like this is dropping a public function of the SDK? Is that a BC

slorello89: Or is this just picked up in an override?

dragonmantank: The diff is funky, that function now exists above `toArray()` (and now just calls `toArray() internally). It's line 176 in the new code.

src/Application/MessagesConfig.php
```diff
@@ -18,6 +18,10 @@ class MessagesConfig
     public function setWebhook($type, $url, $method = null)
     {
         if (!($url instanceof Webhook)) {
+            trigger_error(
```
slorello89, line 21: for all of these - the `$url = new Webhook($url, $method);` is dead code now and should be removed

slorello89: or is `trigger_error` not what I think it is?

dragonmantank: As mentioned it chat, it just raises a warning but doesn't stop execution, so the `$url` creation still happens.

src/Redact/ClientFactory.php
```diff
@@ -0,0 +1,33 @@
...
+        $api = $container->get(APIResource::class);
+        $api
+            ->setBaseUri('/v1/redact/transaction')
+            ->setCollectionName('')
+        ;
+
+        // This API has a slightly different format for the error message, so override
+        $exceptionHandler = $api->getExceptionErrorHandler();
+        if ($exceptionHandler instanceof APIExceptionHandler) {
+            $exceptionHandler->setRfc7807Format("%s - %s. See %s for more information");
```
slorello89, line 27: Is there a reason for making this inconsistent with our other RFC7807 exceptions? The other one uses a `:` instead of a ` -` it seems.

dragonmantank: That’s just how this API is, so it’s the oddball out.

test/Account/ClientTest.php
```diff
@@ -35,11 +38,27 @@ class ClientTest extends TestCase
...
+     */
+    protected $api;
+
     public function setUp()
     {
         $this->nexmoClient = $this->prophesize('Nexmo\Client');
         $this->nexmoClient->getRestUrl()->willReturn('https://rest.nexmo.com');
         $this->nexmoClient->getApiUrl()->willReturn('https://api.nexmo.com');
+
+        // $this->apiClient = new APIResource();
```
slorello89, line 52: looks like you took a more direct route to creating the account client and left a commented out version?

test/Account/responses/prefix-pricing-no-results.json
```diff
@@ -1,4 +1,4 @@
 {
     "count": 0,
-    "countries": []
+    "prices": []
```
slorello89, line 3: I think this should be `countries` shouldn't it?  https://developer.nexmo.com/api/pricing#retrievePrefixPricing

test/Account/ClientTest.php
dragonmantank: That's actually the new "proper" way, so I left it in for demo purposes. v2.2.0 can use either the new way with the `APIResource` to preserve backward-compatibility, or it will build it's own. v3.0.0 makes `APIResource` a hard requirement.

test/Account/responses/prefix-pricing-no-results.json
dragonmantank: Good catch, will fix

test/SMS/Message/VcalTest.php
```diff
@@ -0,0 +1,26 @@
...
+declare(strict_types=1);
+
+namespace NexmoTest\SMS\Message;
+
+use Nexmo\SMS\Message\Vcal;
+use PHPUnit\Framework\TestCase;
+
+class VcalTest extends TestCase
+{
+    public function testCanCreateBinaryMessage()
```
slorello89, line 11: test name looks like copy/paste

test/SMS/Message/VcardTest.php
```diff
@@ -0,0 +1,26 @@
...
+declare(strict_types=1);
+
+namespace NexmoTest\SMS\Message;
+
+use Nexmo\SMS\Message\Vcard;
+use PHPUnit\Framework\TestCase;
+
+class VcardTest extends TestCase
+{
+    public function testCanCreateBinaryMessage()
```
slorello89, line 11: same with this one.

dragonmantank:
For anyone watching this thread, I'll be pulling the trigger and merging 2020-07-28 in the afternoon.

PR #233, opened 2020-07-29
Various fixes to the new Voice client

dragonmantank:
* Allow Conversations to have an `eventUrl` and `eventMethod`
* Back off a bunch of options that have default values but subtly different behavior
* Added missing Notify webhook and new ASR code

### DIFF
--- a/src/Voice/Client.php
+++ b/src/Voice/Client.php
@@ -58,8 +58,13 @@ class Client implements APIClient
             $json['machine_detection'] = $call->getMachineDetection();
         }
 
-        $json['length_timer'] = (string) $call->getLengthTimer();
-        $json['ringing_timer'] = (string) $call->getRingingTimer();
+        if (!is_null($call->getLengthTimer())) {
+            $json['length_timer'] = (string) $call->getLengthTimer();
+        }
+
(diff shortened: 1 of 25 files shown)


src/Voice/Client.php
```diff
@@ -58,8 +58,13 @@ public function createOutboundCall(OutboundCall $call) : Event
             $json['machine_detection'] = $call->getMachineDetection();
         }
 
-        $json['length_timer'] = (string) $call->getLengthTimer();
-        $json['ringing_timer'] = (string) $call->getRingingTimer();
+        if (!is_null($call->getLengthTimer())) {
+            $json['length_timer'] = (string) $call->getLengthTimer();
```
bencgreenberg, line 62: According to the API reference, `length_timer` and `ringing_timer` are Integers, not Strings. Are you treating them as Strings here for an SDK specific reason?

dragonmantank: I don't remember. The API will take both (it's JSON, so it's not a big deal), I just re-tested. It may have been to conserve some backward compatibility with the old Calls client this is replacing.

dragonmantank:
Thanks @bencgreenberg !

PR #236, opened 2020-08-13
Calls Collection/Client now handles being called by MapFactory

dragonmantank:
`MapFactory` now supports factories via the `__invoke()` method on a class, and `Nexmo\Calls\Collection` is of course an invokable object. That didn't play well.

This removes the typehint on `Nexmo\Calls\Collection::__invoke()`, which is fine since this class is deprecated anyway.

### DIFF
--- a/src/Call/Collection.php
+++ b/src/Call/Collection.php
@@ -59,9 +59,10 @@ class Collection implements ClientAwareInterface, CollectionInterface, \ArrayAcc
      * @param null $callOrFilter
      * @return $this|Call
      */
-    public function __invoke(Filter $filter = null)
+    public function __invoke($filter = null)
     {
-        if (!is_null($filter)) {
+        /** Fix for the smarter MapFactory in v2.2.0 and the uniqueness of this class interface */
+        if (!is_null($filter) && $filter instanceof Filter) {
             $this->setFilter($filter);
         }
 


dragonmantank:
Testing:

1. Make a new directory
2. Run `composer require nexmo/client-core:dev-bug/235-calls-collection-invoke@dev php-http/guzzle6-adapter`
3. Create the following test script as `test.php`
4. Run with `TO_NUMBER=<your_number> FROM_NUMBER=<nexmo_number> KEY_PATH=</path/to/key> APP_ID=<application_id> php test.php`

You should get a telephone call with a TTS message.

```php
<?php

use Nexmo\Client\Credentials\Keypair;

require_once __DIR__ . '/vendor/autoload.php';

define('KEY_PATH', getenv('KEY_PATH');
define('APP_ID', getenv('APP_ID');
define('TO_NUMBER', getenv('TO_NUMBER');
define('FROM_NUMBER', getenv('FROM_NUMBER');

$client = new Nexmo\Client(
    new Keypair(
        file_get_contents(KEY_PATH),
        APP_ID
    )
);

$call = $client->calls()->create([
    'to' => [[
        'type' => 'phone',
        'number' => TO_NUMBER
    ]],
    'from' => [
        'type' => 'phone',
        'number' => FROM_NUMBER
    ],
    'answer_url' => ['https://developer.nexmo.com/ncco/tts.json'],
]);```

slorello89:
LGTM 👍

PR #240, opened 2020-08-19
Added support for the PSD2 endpoint

dragonmantank:
This adds support for the PSD2 endpoint for Verify, and a new Value Object to handle the request. This falls in line with what v3.0.0's Verify interface will look like.

```php
$request = new \Nexmo\Verify\RequestPSD2('12225551234', 'Payee', '9.99');
$response = $nexmo->verify()->requestPSD2($request);
echo $response['request_id'];
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,13 +164,15 @@ Using your signature secret and the other supplied parameters, the signature can
 Vonage's [Verify API][doc_verify] makes it easy to prove that a user has provided their own phone number during signup,
 or implement second factor authentication during signin.
 
-You can start a verification process using a simple array:
+You can start a verification process using code like this:
 
 ```php
 $request = new \Vonage\Verify\Request('14845551212', 'My App');
 $response = $client->verify()->start($request);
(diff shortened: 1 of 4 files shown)


lornajane:
Gonna rebase this one as well and try it with the Vonage namespace and no conflict ...

lornajane (commented):
Woot! So happy to have PSD2 support and this feature looks pretty close :)
 - I rebased to get the Vonage namespace change and avoid a merge conflict (~I got a build failure but only for 7.1, I don't know why? Does it handle the bridge stuff differently?~ Never mind, I fixed what I broke!)
 - The vanilla verify request returns an object. This returns an array. It feels a bit inconsistent but I'm sure there are reasons! Is the array our preferred thing now?
 - I added README docs, so can you take a look and I don't think you can review your own PR but if you're happy then I can approve + merge.

dragonmantank:
Can't review but I'm allowed to always merge :P

PR #242, opened 2020-08-27
Fixed passing through and defaulting to client API URL

dragonmantank:
By default, the `\Vonage\Client\APIResource` would use the default Base API URL from the `\Vonage\Client::BASE_API` constant. If the user override the API URL, this was never passed through to the `\Vonage\Client\APIResource`. Rest URL overrides would be, because most of the factories for APIs hitting `rest.nexmo.com` would pass this up manually.

Defaulting to the API URL was because most of the APIs use `api.nexmo.com`, and `rest.nexmo.com` is only handling a few legacy APIs.

This fix makes it so the `\Vonage\Client\APIResource` defaults to a blank URL, and if a client is assigned to grab the client's API URL. Otherwise it returns a blank, and the underling HTTP Client can deal with it.

This also includes a minor fix where setting the Exception Handler on the API resource didn't return `$this`, breaking chaining.

### DIFF
--- a/src/Client/APIResource.php
+++ b/src/Client/APIResource.php
@@ -19,7 +19,7 @@ class APIResource implements ClientAwareInterface
      * client or directly on this class.
      * @var string
      */
-    protected $baseUrl = Client::BASE_API;
+    protected $baseUrl = '';
 
     /**
      * @var string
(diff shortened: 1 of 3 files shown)


lornajane (approved):
This behaves as I expected, thanks!

PR #248, opened 2020-09-02
Added make() method to bypass service locator caching

dragonmantank:
## Description

The `\Vonage\Client\Factory\MapFactory` originally worked by using a cache to cut down on object instantiation. This adds a `make()` method so that the cache can be bypassed and the object pulled from the `MapFactory` is always instantiated fresh.

This also includes a few tweaks to the Account Client to make it easier to test, and a typo-fix for the Account ClientFactory.

## Motivation and Context

@slorello89 noticed that with the Secrets API, it would try and use `rest.nexmo.com` instead of `api.nexmo.com`. This was because the `\Vonage\Accounts\ClientFactory` generates two `\Vonage\Client\APIResource` objects, as Secrets and Accounts have two different base URLs.

Because PHP always uses references when passing around objects, the `MapFactory` instance was updated to use `rest.nexmo.com` instead of the default `api.nexmo.com`.

While I could just fix the `ClientFactory`, that doesn't solve the underlying problem if multiple APIs are used in a single script.

## How Has This Been Tested?

Locally against the code snippet

## Types of changes

## Checklist:

### DIFF
--- a/src/Account/Client.php
+++ b/src/Account/Client.php
@@ -41,7 +41,7 @@ class Client implements ClientAwareInterface, APIClient
      * Shim to handle older instatiations of this class
      * @deprecated Will remove in v3
      */
-    protected function getAccountAPI() : APIResource
+    public function getAccountAPI() : APIResource
     {
         if (is_null($this->accountAPI)) {
             $api = new APIResource();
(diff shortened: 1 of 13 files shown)


slorello89 (approved):
LGTM

PR #250, opened 2020-09-09
Bump minimum PHP version to 7.2

dragonmantank:
## Description

* Package now requires PHP 7.2 or higher
* Changed dev dependency from Guzzle Adapter to straight Guzzle 7
* Support either the Guzzle 6 HTTPlug adapter _or_ Guzzle 7 automatically if the classes exist
* Dropped PHP 7.1 from testing matrix
* Updated docs

## Motivation and Context

Laravel 8 was released, which now ships with and recommends upgrading to Guzzle 7. This impacts our `nexmo/laravel` package as it requires Guzzle 6. While there is a workaround, this still causes friction with users. Updating the core library makes it easier to update the Laravel package.

PHP 7.1 is also EOL, and only represents 3-4% of API usage on Vonage, with PHP 7.2+ representing ~73% of usage (the resulting ~27% are 7.1 or lower versions that we no longer support officially). 

## How Has This Been Tested?

Tested against a few code snippets, since all the code snippets use the `\Vonage\Client` object and whatever HTTP client it's configured to (in this case, Guzzle 7 by default).

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     "email": "devrel@nexmo.com"
   },
   "require": {
-    "php": ">=7.1",
+    "php": ">=7.2",
     "psr/http-client-implementation": "^1.0",
     "laminas/laminas-diactoros": "^1.8.4 || ^2.0",
     "lcobucci/jwt": "^3.2",
@@ -43,7 +43,7 @@
     "php-http/mock-client": "^1.4",
     "estahn/phpunit-json-assertions": "^3.0.0",
     "squizlabs/php_codesniffer": "^3.1",
-    "php-http/guzzle6-adapter": "^2.0",
+    "guzzlehttp/guzzle": "^7.0",
     "phpstan/phpstan": "^0.12.11"
   },
   "autoload": {
--- a/src/Client.php
+++ b/src/Client.php
@@ -77,7 +77,13 @@ class Client
     public function __construct(CredentialsInterface $credentials, $options = array(), ClientInterface $client = null)
     {
         if (is_null($client)) {
-            $client = new \Http\Adapter\Guzzle6\Client();
+            // Since the user did not pass a client, try and make a client
+            // using the Guzzle 6 adapter or Guzzle 7
+            if (class_exists(\Http\Adapter\Guzzle6\Client::class)) {
+                $client = new \Http\Adapter\Guzzle6\Client();
+            } elseif (class_exists(\GuzzleHttp\Client::class)) {
+                $client = new \GuzzleHttp\Client();
+            }
         }
 
         $this->setHttpClient($client);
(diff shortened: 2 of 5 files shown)


src/Client.php
```diff
@@ -77,7 +77,11 @@ class Client
     public function __construct(CredentialsInterface $credentials, $options = array(), ClientInterface $client = null)
     {
         if (is_null($client)) {
-            $client = new \Http\Adapter\Guzzle6\Client();
+            if (class_exists(\Http\Adapter\Guzzle6\Client::class)) {
+                $client = new \Http\Adapter\Guzzle6\Client();
```
lornajane, line 81: Why is this here if we've specified Guzzle 7 in the dependencies? Please add a code comment because I'll be confused next time I see this too!

lornajane (commented):
I'm on board with bumping the PHP requirement, especially since even 7.2 is barely a stable version https://www.php.net/supported-versions.php. I had a question but it's minor.

src/Client.php
dragonmantank: If someone is already on 7.2+ and using the Guzzle 6 adapter, we're keeping the magic "hey this works", and extending it to also check for Guzzle 7.

dragonmantank: Updated with a comment

composer.json
```diff
@@ -43,7 +43,7 @@
     "php-http/mock-client": "^1.4",
     "estahn/phpunit-json-assertions": "^3.0.0",
     "squizlabs/php_codesniffer": "^3.1",
-    "php-http/guzzle6-adapter": "^2.0",
+    "guzzlehttp/guzzle": "^7.0",
```
slorello89, line 46: I'm not too familiar with what's going on here - but from the description, it sounds like we should be pointing at `^6.0` - why the slightly higher major version?

slorello89: (because Laravel is jumping to 6.0)

dragonmantank: It's actually two completely different packages. `php-http/guzzle6-adapter` was a wrapper library for Guzzle 6, which made it PSR-18 compatible (a spec for HTTP clients that make them framework-compatible). By targeting PSR-18 we can support a range of HTTP clients instead of just one (Guzzle). So this is pinned against v2 of the wrapper, which wrapped v6 of Guzzle.

Guzzle 7 is PSR-18 compatible so there isn't a wrapper, and we can use Guzzle 7 directly as a development dependency.

src/Client.php
lornajane: makes sense. Why prefer the v6 adapter over the v7 if it is present? (sorry, I realise you want this merged)

dragonmantank: No specific reasoning other than that was our previous option, and I don't want to break any existing users in a minor release. With v3 I'll look at just auto-creating a Guzzle 7 object

lornajane (approved):
Calling it good!

PR #257, opened 2020-10-07
2.6.0

SoftCreatR:
* Dropped support for PHPUnit 7
* Added missing PHPDoc blocks
* Added missing return type hints
* Replaced qualifiers with imports
* Updated and optimized examples
* Applied multiple code optimizations (especially for PHP 7.2+) and simplified some logic
* Updated code styling to match PSR standards
* Re-ordered imports where necessary
* Updated tests to get rid of deprecation messages
* Fixed namespace declarations in tests
* Updated phpunit.xml.dist
* Added Roave Security Advisories as dev-requirement to prevent usage of packages with known security vulnerabilities
* Replaced estahn/phpunit-json-assertions with martin-helmich/phpunit-json-assert due do compatibility issues with PHPUnit
* Removed test build for PHP 7.1 in .travis.yml
* Added missing punctuation in CONTRIBUTING.md
* Updated contact email address in CODE_OF_CONDUCT.md


SoftCreatR:
@Lczajkowski I'm still not done 😄

SoftCreatR:
Finally 🚀 🔥 🥳 🎉

composer.json
```diff
@@ -1,14 +1,16 @@
 {
   "name": "vonage/client-core",
+  "type": "library",
   "description": "PHP Client for using Vonage's API.",
+  "homepage": "https://www.vonage.com",
```
dragonmantank, line 5: I would set this to `https://developer.nexmo.com` so that it points people directly to the docs

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: php
```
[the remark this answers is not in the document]
SoftCreatR: I've just refactored what's there and this file exists in your repo. But sure,i can just delete it.

composer.json
SoftCreatR: But that's not the purpose of this field. For this, there's a field called `support` and there, the docs are linked.

dragonmantank (changes requested):
Did a quick spot check for some high-level items, so this is not a comprehensive review. 

Overall looks really good, but have some questions and some reverts and changes I would like to see done (mainly based on the why we did some things why we did).

I will have more as I chunk this up and dig through.

composer.json
dragonmantank: We consider it the homepage for our APIs though, even though it's not the company homepage.

.travis.yml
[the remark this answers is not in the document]
dragonmantank: Totally understand! We switched to Github Actions and never deleted this file in case we needed to switch back, but GHA has been working fine. I just hadn't gotten around to deleting it myself :P

dragonmantank:
Marking this accepted ... there's a lot of good changes in here so thanks for doing a lot of this work! Just want to make sure you get both Hacktoberfest and Vonage credit for the work in case we don't merge it in time

src/Account/ClientFactory.php
```diff
@@ -1,22 +1,31 @@
...
+    public function __invoke(ContainerInterface $container): Client
     {
         /** @var APIResource $accountApi */
         $accountApi = $container->make(APIResource::class);
         $accountApi
             ->setBaseUrl($accountApi->getClient()->getRestUrl())
             ->setIsHAL(false)
-            ->setBaseUri('/account')
-        ;
+            ->setBaseUri('/account');
```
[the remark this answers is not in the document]
SoftCreatR: > Does that cause a conflict with one of the rulesets?

Not really. But most IDEs will probably remove it.

src/Account/Price.php
```diff
@@ -142,52 +185,85 @@ public function fromArray(array $data)
...
         if ($key === 'data') {
             trigger_error(
                 "Direct access to " . get_class($this) . "::data is deprecated, please use getter to toArray() methods",
                 E_USER_DEPRECATED
             );
+
             return $this->data;
         }
+
+        return [];
```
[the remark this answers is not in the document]
SoftCreatR: Wouldn't `mixed` fit better? Feels like cheating, but works as expected.

src/Application/Application.php
```diff
@@ -103,16 +141,22 @@ public function getMessagesConfig()
...
+     * @throws Exception
      */
     public function getRtcConfig()
     {
         if (!isset($this->rtcConfig)) {
             $this->setRtcConfig(new RtcConfig());
             $data = $this->getResponseData();
-            if (isset($data['rtc']) and isset($data['rtc']['webhooks'])) {
+
+            if (isset($data['rtc']['webhooks'])) {
```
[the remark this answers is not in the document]
SoftCreatR: That's not true for `empty` :)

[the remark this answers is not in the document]
SoftCreatR: http://sandbox.onlinephpfunctions.com/code/3b7370c9c80919e4e7fe1313f7ac9a12879d9b58

src/Application/Application.php
```diff
@@ -173,44 +235,53 @@ public function jsonUnserialize(array $json)
...
+        $this->id = $data['id'] ?? null;
         $this->keys = $data['keys'] ?? [];
 
         if (isset($data['capabilities'])) {
             $capabilities = $data['capabilities'];
 
             //todo: make voice  hydrate-able
             $this->voiceConfig = new VoiceConfig();
-            if (isset($capabilities['voice']) and isset($capabilities['voice']['webhooks'])) {
+            if (isset($capabilities['voice']['webhooks'])) {
```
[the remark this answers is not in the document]
SoftCreatR: http://sandbox.onlinephpfunctions.com/code/3b7370c9c80919e4e7fe1313f7ac9a12879d9b58

src/Account/Price.php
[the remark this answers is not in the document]
dragonmantank: `mixed` would be a good return docblock value since we don't really know what the return value is, and will cover `null`.

A lot of the array access code is being nuked in v3, so this just avoids a BC break. If we start returning `[]` instead of `null` we have the possibility of messing up a developer's checks.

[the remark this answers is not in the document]
SoftCreatR: Well, `mixed` wouldn't fit. Instead, I'll hint it as `array|null`, since $data is (and was) annotated as an array (which is, why i returned an empty array instead of `null`).

src/Application/Application.php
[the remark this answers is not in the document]
dragonmantank: I stand corrected!

Though if you do see any `and` conditionals please swap them to `&&` for consistency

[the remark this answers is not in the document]
SoftCreatR: That's what I've done, where needed ;)

src/Account/ClientFactory.php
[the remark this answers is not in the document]
SoftCreatR: I would leave this for now, since there are many places where method chaining is used, and my IDE keeps trimming those lines.

src/Account/Config.php
```diff
@@ -20,51 +28,88 @@ class Config implements
      */
     protected $data = [];
 
-    public function __construct($sms_callback_url = null, $dr_callback_url = null, $max_outbound_request = null, $max_inbound_request = null, $max_calls_per_second = null)
-    {
+    /**
+     * Config constructor.
+     *
+     * @param null $sms_callback_url
```
dragonmantank, line 34: I would set all these to `string|null`

dragonmantank (changes requested):
Handful of other cleanup items, currently about 75 files total into the PR :P

src/Client.php
```diff
@@ -119,21 +134,21 @@ public function __construct(CredentialsInterface $credentials, $options = array(
 
         $this->setFactory(new MapFactory([
             // Legacy Namespaces
-            'message' => \Vonage\Message\Client::class,
-            'calls' => \Vonage\Call\Collection::class,
-            'conversation' => \Vonage\Conversations\Collection::class,
-            'user' => \Vonage\User\Collection::class,
+            'message' => Message\Client::class,
```
dragonmantank, line 137: For consistency, I would either revert this or rename them all the `[X]Client` via the `use ... as [X]Client`

src/Client.php
```diff
@@ -142,26 +157,26 @@ public function __construct(CredentialsInterface $credentials, $options = array(
         // Disable throwing E_USER_DEPRECATED notices by default, the user can turn it on during development
         if (array_key_exists('show_deprecations', $this->options) && !$this->options['show_deprecations']) {
             set_error_handler(
-                function (
```
dragonmantank, line 145: While we don't use these parameters, I would keep them so they match the PHP docs

src/Client.php
```diff
@@ -172,19 +187,20 @@ public function getApiUrl()
...
         $this->client = $client;
+
         return $this;
     }
 
     /**
      * Get the Http Client used to make API requests.
      *
-     * @return HttpClient
+     * @return \GuzzleHttp\Client|HttpClient
```
dragonmantank, line 203: This could just be `ClientInterface` now, since we now support PSR-18 directly versus our wrapper we used to have

src/Client.php
```diff
@@ -406,21 +450,27 @@ public function delete($url)
...
+     * @return ResponseInterface
+     * @throws ClientExceptionInterface
+     * @throws Exception
+     */
+    public function send(RequestInterface $request): ResponseInterface
     {
         if ($this->credentials instanceof Container) {
             if ($this->needsKeypairAuthentication($request)) {
-                $request = $request->withHeader('Authorization', 'Bearer ' . $this->credentials->get(Keypair::class)->generateJwt());
+                $c = $this->credentials->get(Keypair::class)->generateJwt();
```
dragonmantank, line 464: I would rename `$c` -> `$token` for clarity, since we're getting a JWT out of this call.

src/Client.php
```diff
@@ -406,21 +450,27 @@ public function delete($url)
...
+
+                $request = $request->withHeader('Authorization', 'Bearer ' . $c);
             } else {
                 $request = self::authRequest($request, $this->credentials->get(Basic::class));
             }
         } elseif ($this->credentials instanceof Keypair) {
-            $request = $request->withHeader('Authorization', 'Bearer ' . $this->credentials->generateJwt());
+            $c = $this->credentials->generateJwt();
```
dragonmantank, line 471: Same here, `$c` -> `$token`

src/Client.php
```diff
@@ -447,66 +497,82 @@ public function send(\Psr\Http\Message\RequestInterface $request)
...
-            $userAgent[] = $app['name'].'/'.$app['version'];
+            $userAgent[] = $app['name'] . '/' . $app['version'];
         }
 
         // Set the header. Build by joining all the parts we have with a space
         $request = $request->withHeader('User-Agent', implode(" ", $userAgent));
 
-        $response = $this->client->sendRequest($request);
-        return $response;
+        return $this->client->sendRequest($request);
```
dragonmantank, line 514: This is going to sound dumb, but keeping this separate helps a lot with debugging. There's a lot of times I have to dump out or inspect the response, so having it as a bespoke variable makes that easier.

src/Client.php
```diff
@@ -518,16 +584,25 @@ public function __call($name, $args)
...
+    /**
+     * @param $name
+     * @return mixed
+     * @noinspection MagicMethodsValidityInspection
+     */
     public function __get($name)
     {
-        if (!$this->factory->has($name)) {
-            throw new \RuntimeException('no api namespace found: ' . $name);
+        if (!$this->factory->hasApi($name)) {
```
dragonmantank, line 594: This should be `has`, which is PSR-11 compliant.

src/Client.php
```diff
@@ -518,16 +584,25 @@ public function __call($name, $args)
...
+            throw new RuntimeException('no api namespace found: ' . $name);
         }
 
-        return $this->factory->get($name);
+        return $this->factory->getApi($name);
```
dragonmantank, line 598: This should be `get`, which is PSR-11 compliant.

src/Call/Talk.php
```diff
@@ -135,34 +193,52 @@ protected function getException(ResponseInterface $response)
...
-    public function offsetSet($offset, $value)
+    /**
+     * @param mixed $offset
+     * @param mixed $value
+     */
+    public function offsetSet($offset, $value): void
     {
-        if (!in_array($offset, $this->params)) {
-            throw new \RuntimeException('invalid parameter: ' . $offset);
+        if (!in_array($offset, $this->params, false)) {
```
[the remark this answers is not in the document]
SoftCreatR: Old habits, I guess.

src/Client/ClientAwareTrait.php
```diff
@@ -17,18 +19,26 @@ trait ClientAwareTrait
...
+     * @return Client|null
+     */
+    public function getClient(): ?Client
     {
         if (isset($this->client)) {
             return $this->client;
         }
 
-        throw new \RuntimeException('Vonage\Client not set');
+        throw new RuntimeException('Vonage\Client not set');
```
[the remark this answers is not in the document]
SoftCreatR: But it was there already?

src/Call/Talk.php
```diff
@@ -54,27 +71,47 @@ public function __invoke(self $entity = null)
         return $this->put($entity);
     }
 
+    /**
+     * @return mixed
```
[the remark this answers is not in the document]
SoftCreatR: Must be `string|null` then :)

src/Client/APIResource.php
```diff
@@ -63,7 +70,15 @@ class APIResource implements ClientAwareInterface
      */
     protected $lastResponse;
 
-    public function create(array $body, string $uri = '', $headers = []) : ?array
+    /**
+     * @param array $body
```
dragonmantank, line 74: To help with visual debt, I would remove any docblocks where we have clear type hints. It's less maintenance.

dragonmantank (changes requested):
Handful of new changes

dragonmantank:
To expand on my "visual debt" note:

Visual Debt is anything that adds a cognitive load on the developer. This can come in many different forms, especially extraneous comments or complex methods. The harder a developer has to think to understand a block of code, the worse the code is. Another side of this is as the user has to _read_ more to understand the code, the harder time a developer has understanding the code. 

A lot of this code is very hold, so there are areas where there aren't any docblocks or type hints, some areas where we have docblocks but no hints, and somewhere we have both, and some where we only have type hints. It's a weird archeological dig through the history of PHP and the things that were needed at the time.

With the addition of type hinting, we can clear out one layer of visual debt - `@param` and `@return` annotations. Traditionally they existed to help IDEs and developers know what type of data was expected, but PHP can now directly enforce that with type hints. By duplicating the information as both a type hint and an annotation, it's not only two places for maintainers to keep up-to-date, but it's also additional information for developers to check.

As a good rule of thumb, if we have type hints, we don't need `@param` or `@return`. 9 times out of 10 these work well, with the other times being arrays that we can clarify (https://psalm.dev/docs/annotating_code/type_syntax/array_types/) or where we need to add clarification on _why_ we take a data type. 

Docblocks and comments exist to provide clarity on the code, not just to simply document it exists. 

Docblocks for things like `@throws` are still good, we have no code-equivalent way of documenting that up to an IDE. 

I use the following rules:
- If the method name is clear, I don't need to document the idea behind the method
- If the parameter has a type hint, I don't need to document the parameter
- If the parameter takes a clear, concrete type, type hint over docblocking
- If the parameter takes multiple types, docblock those types and leave off the type hint
- If the return value is type hinted, I don't need to document the return type
- If the return value is a clear, concrete type, type hint over docblocking
- If the return value is ambiguous or mixed, docblock those types and leave off the type hint

src/Client.php
SoftCreatR: Then, tests fail:

> Prophecy\Exception\Doubler\MethodNotFoundException : Method `Double\FactoryInterface\P4::has()` not found.

src/Client.php
SoftCreatR: Then, tests fail:

> Prophecy\Exception\Doubler\MethodNotFoundException : Method `Double\FactoryInterface\P4::get()` not found.

SoftCreatR:
While I understand most of your explanation, I do not fully agree with all points. At the end (v2.0+) you would only use doc blocks if you have something to explain and omit them in all other cases, which makes a Developer think, that every unannotated param and every unannotated return type just means `mixed`. Dunno, if that's really the expected behavior.

However, do you want me to remove all doc blocks, that only describe params and return values?

dragonmantank:
> which makes a Developer think, that every unannotated param and every unannotated return type just means `mixed`

Not quite. A developer has no idea if we provide _no_ documentation (annotations, type hints, or comments). We should provide the best and most appropriate information we can though, which for parameters and return values can be annotations _or_ type hints. There's no need for both.

## 1. Worst Case

```php
function sendSMS($to, $from, $message) {}
```

This would be the worst case for a function (or method). We're providing no direction on what those variables are or their data types. The developer could make a reasonable inference that `$to` and `$from` are telephone numbers, and that `$message` is a string message. But for `$to` and `$from` should they pass an integer, a string, or an object? It's not clear, so the developer will have to dig into the function to see what it expects. 

The developer also has no idea if that function returns anything without looking at the source code.

## 2. Better
```php
/**
 * @param string $to
 * @param string $from
 * @param string $message
 * @return array $response
 */
function sendSMS($to, $from, $message) {}
```

This is better. The developer knows that each of the parameters are strings, and that the return value is some sort of an array. 



## 3. Even Better
```php
function sendSMS(string $to, string $from, string $message): array {}
```
Like the previous version, this gives the developer a bunch of information about the function. This is better because not only does it provide the exact same information as the annotations, it's actually enforceable at the engine level. 

## 4. Best
```php
/**
 * @throws ThrottleException
 * @throws ResponseException
 * @return array{status: integer, messageCount: integer, messages: array{price: string, messageId: string}}
 */
function sendSMS(string $to, string $from, string $message): array {}
```

This is the best case. Not only are we letting the developer know what exceptions may happen, but we're also documenting the return array (if this was a string or bool or more basic scalar, we could leave this annotation off). They no longer need to dig into any other source code to figure out how to handle the response, and can wrap their code more appropriately. 

<hr/>

> However, do you want me to remove all doc blocks, that only describe params and return values?

If we already have type hints, the `@param` and `@return` annotations do nothing but duplicate information the developer (or tool) already has (except arrays because PHP dropped the ball on that one). 

So when it comes to refactoring, if we are at \#1, we should move to \#2 by providing basic annotations. This is the safest bet as it avoids any errors with strict typing or type handling. The code will continue to work like it always has, with the additional documentation. While I didn't document it above, using the annotations from \#2 and \#4 gives you the best documentation-only developer docs.

If we are already at \#2, we can move to \#3 or \#4. #3 alone will get us safer code, and #4 will provide additional documentation. 

So if no type hints and no annotations exist, we should add annotations. If type hints exist, only annotate what is needed.

src/Client/ClientAwareTrait.php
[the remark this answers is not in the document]
dragonmantank: It was, blindness on my part :)

SoftCreatR:
@dragonmantank What about this?

https://github.com/Vonage/vonage-php-sdk-core/pull/257/commits/e9fc64e7631e022a60db69846b60fa6a0d0f1ca2

However, I'm not sure, why the test fails now :/

SoftCreatR:
@dragonmantank What do you think? Should I just close this PR and create a new one, based on v3.0.0? For some reason, I think that'd be more useful.

dragonmantank:
I'll take a look at the tests here in a bit. We can finish this PR out, I tried updating the current v3 branch and the rebrand really messed it up. So since this is BC compatible, we'll get this one going, and I'll redo v3.

dragonmantank:
```
1) Vonage\Test\Call\CollectionTest::testHydrateSetsDataAndClient
TypeError: Return value of Double\Vonage\Call\Call\P2::setClient() must be an instance of Vonage\Call\Call, null returned

/home/ctankersley/Projects/vonage/php/vonage-php-sdk-core/src/Call/Collection.php:69
/home/ctankersley/Projects/vonage/php/vonage-php-sdk-core/test/Call/CollectionTest.php:80
```

I'm not sure why that double thinks it should be returning a Call object, it may be an issue with newer versions of the mocking library. 

Looking at that test though, it's just a bad test overall. It's testing to make sure methods are called but not that the end result even looks correct. 

Just mark that test Incomplete:

```php
$this->markTestIncomplete('Rework this test to make sure hydrate fills correct values');
```

dragonmantank:
Overall the changes look cleaner as well, doing some spot checks.

SoftCreatR:
> Overall the changes look cleaner as well, doing some spot checks.

If you are fine with the latest changes on the doc blocks, I'll do them in all source files.

dragonmantank:
Go ahead!

SoftCreatR:
Removing all the doc blocks that I've added takes longer, than expected 🤣

test/Voice/Webhook/FactoryTest.php
```diff
@@ -1,238 +1,285 @@
...
 use Vonage\Voice\Webhook\Transfer;
-use Zend\Diactoros\ServerRequest;
-use Vonage\Voice\Webhook\Notification;
-use Zend\Diactoros\Request\Serializer;
 
 class FactoryTest extends TestCase
 {
-    public function testCanGenerateStartedEvent()
+    /**
+     * @throws Exception
```
dragonmantank, line 31: These `@throws Exception` call outs for the tests are unnecessary. Users should never be invoking these tests in their code.

test/Voice/Webhook/FactoryTest.php
```diff
@@ -1,238 +1,285 @@
...
+/**
+ * Vonage Client Library for PHP
+ *
+ * @copyright Copyright (c) 2016-2020 Vonage, Inc. (http://vonage.com)
+ * @license https://github.com/Vonage/vonage-php-sdk-core/blob/master/LICENSE.txt Apache License 2.0
+ */
 declare(strict_types=1);
 
-namespace VonageTest\Voice;
+namespace Vonage\Test\Voice\Webhook;
```
dragonmantank, line 10: I'd like to revert this. I know some frameworks prefer to keep tests in the same namespace, but I prefer keeping the test namespace completely separate from the core namespace. This helps avoid collisions where maybe we want a bespoke `Vonage\Test` namespace in the future, but makes it much more clear this is not user-invokable code.

src/Message/AutoDetect.php
```diff
@@ -13,26 +14,27 @@
...
      * Message Body
      * @var string
      */
     protected $text;
-    
+
     /**
      * Create a new SMS text message.
      *
-     * @param string $to
```
dragonmantank, line 27: I would revert this, as the new annotations are less useful

src/Message/Binary.php
```diff
@@ -2,54 +2,62 @@
...
-    
+
     /**
      * Create a new SMS text message.
      *
-     * @param string $to
-     * @param string $from
-     * @param string $body
-     * @param string $udh
+     * @param $to
```
dragonmantank, line 36: I would revert these as well, since it's removing info

src/Message/Client.php
```diff
@@ -255,10 +290,12 @@ public function get($query)
 
     /**
      * @todo Fix all this error detection so it's standard
+     * @param $idOrMessage
+     * @return InboundMessage|Message|MessageInterface|null
+     * @throws ClientExceptionInterface
+     * @throws Exception\Exception
+     * @throws Exception\Request
      * @deprecated Please use the Reports API instead
-     * @param string|MessageInterface $idOrMessage
```
dragonmantank, line 259: I would revert this as well, since it's removing info

src/Message/EncodingDetector.php
```diff
@@ -2,33 +2,167 @@
...
-                '@', '£', '$', '¥', 'è', 'é', 'ù', 'ì', 'ò', 'ç', "\r", 'Ø', 'ø', "\n", 'Å', 'å',
-                'Δ', '_', 'Φ', 'Γ', 'Λ', 'Ω', 'Π', 'Ψ', 'Σ', 'Θ', 'Ξ', 'Æ', 'æ', 'ß', 'É',
-                ' ', '!', '"', '#', '¤', '%', '&', '\'', '(', ')', '*', '+', ',', '-', '.', '/',
-                '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', ':', ';', '<', '=', '>', '?',
-                '¡', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O',
-                'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 'Ä', 'Ö', 'Ñ', 'Ü', '§',
-                '¿', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o',
-                'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 'ä', 'ö', 'ñ', 'ü', 'à',
-                "\f", '^', '{', '}', '\\', '[', '~', ']', '|', '€',
+            [
```
dragonmantank, line 23: I would revert this, it's making the file unnecessarily long and it's not going to improve any readability. This list also will never change, so not super worried about needing to maintain the list.

src/Message/EncodingDetector.php
```diff
@@ -40,12 +174,15 @@ public function requiresUnicodeEncoding($content)
...
     }
 
-    private function convertIntoUnicode()
+    /**
+     * @return callable
+     */
+    private function convertIntoUnicode(): callable
     {
-        return function ($char) {
+        return static function ($char) {
```
dragonmantank, line 182: Is there a reason this is being converted to a `static` function? It's just a callable, and doesn't hold any persistent data inside of itself.

src/Message/Message.php
```diff
@@ -317,20 +417,27 @@ public function offsetExists($offset)
...
+     */
     public function offsetGet($offset)
     {
         trigger_error(
             "Array access for " . get_class($this) . " is deprecated, please use getter methods",
             E_USER_DEPRECATED
         );
-        $response = @$this->getResponseData();
+
+        $response = $this->getResponseData();
```
dragonmantank, line 433: I would put this (and other silencers around request/response data) silencer back in.

It's silenced to help reduce the amount of deprecation warnings that happen on things the user can't change. For example, iterating over the Message object has the potential to generate a ton of deprecation warnings not only for the messages, but has a side effect of the response/request data access also being deprecated and triggering their own warnings, and it causes a cascading effect.

src/Message/Response/Collection.php
```diff
@@ -27,43 +30,53 @@ class Collection extends Response implements ResponseInterface, \Countable, \Ite
...
+    /**
+     * Collection constructor.
+     *
+     * @param array $data
+     */
     public function __construct(array $data)
     {
-        $this->expected = array('message-count', 'messages');
-        $return = parent::__construct($data);
+        parent::__construct($data);
```
dragonmantank, line 47: These need swapped around, as the parent constructor relies on `$this->expected` being populated.

The other alternative is to just set it as a class property, this code is old enough I don't know if there was a reason we didn't. Either way, that array needs populated before calling the parent.

src/Message/Response/Message.php
```diff
@@ -21,101 +22,94 @@ class Message extends Response implements ResponseInterface
...
             'message-price',
             'network'
-        );
+        ];
 
         //default value
-        $data = array_merge(array('client-ref' => null, 'remaining-balance' => null), $data);
+        $data = array_merge(['client-ref' => null, 'remaining-balance' => null], $data);
 
-        $return = parent::__construct($data);
```
dragonmantank, line 35: Erring on the side of caution, I'd revert all the logic in this constructor but keep any syntactical changes. The logic here is extremely suspect, but this entire namespace is going away in v3 and this logic isn't replicated.

While yes, a `__construct()` method called magically during object creation returns the object, the `__construct()` method is directly callable as well and _can_ return something. With how this is set up, I would rather keep the sketchy behavior in case someone is actively abusing that ability. 

https://3v4l.org/RQPPE

src/Message/Text.php
```diff
@@ -13,25 +14,26 @@
...
      * Message Body
      * @var string
      */
     protected $text;
-    
+
     /**
      * Create a new SMS text message.
      *
-     * @param string $to
```
dragonmantank, line 27: I would revert these so we get the types back

src/Message/Unicode.php
```diff
@@ -2,45 +2,53 @@
...
      * Message Body
      * @var string
      */
     protected $text;
-    
+
     /**
      * Create a new SMS text message.
      *
-     * @param string $to
```
dragonmantank, line 27: I would revert these so we get the types back

src/Message/Vcal.php
```diff
@@ -2,45 +2,51 @@
...
      * Message Body
      * @var string
      */
     protected $vcal;
-    
+
     /**
      * Create a new SMS text message.
      *
-     * @param string $to
```
dragonmantank, line 27: I would revert these so we get the types back

src/Message/Vcard.php
```diff
@@ -24,23 +27,26 @@ class Vcard extends Message
     /**
      * Create a new SMS text message.
      *
-     * @param string $to
```
dragonmantank, line 27: I would revert these so we get the types back

src/Message/Wap.php
```diff
@@ -2,63 +2,71 @@
...
 
     /**
      * Create a new SMS text message.
      *
-     * @param string $to
-     * @param string $from
-     * @param string $title
-     * @param string $url
-     * @param int $validity
+     * @param $to
```
dragonmantank, line 42: I would revert these so we get the types back

dragonmantank (changes requested):
Another round of changes!

We're getting there!

src/Message/Response/Message.php
SoftCreatR: I know, that you COULD use a return value from a constructor. But tbh, that is highly uncommon and fairly not recommended. However, in v3, this should be avoided.

src/Message/EncodingDetector.php
SoftCreatR: Better scoping; in some cases can improve performance.

src/Client.php
```diff
@@ -447,66 +519,84 @@ public function send(\Psr\Http\Message\RequestInterface $request)
...
-    public function unserialize($entity)
+    /**
+     * @param $entity
+     * @return Verification
+     */
+    public function unserialize($entity): Verification
     {
         if (is_string($entity)) {
-            $entity = unserialize($entity);
+            $entity = unserialize($entity, [Verification::class]);
```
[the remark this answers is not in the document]
SoftCreatR: > I think the second parameter has to use an allowed_classes key:

Either an array of class names that should be accepted, FALSE to accept no classes, or TRUE to accept all classes. If this option is defined and unserialize() encounters an object of a class that isn't to be accepted, then the object will be instantiated as __PHP_Incomplete_Class instead. Omitting this option is the same as defining it as TRUE: PHP will attempt to instantiate objects of any class.

test/Voice/Webhook/FactoryTest.php
SoftCreatR: While this is correct, I would just leave them there. They don't hurt and your IDE may thank you later :) And whenever you watch at your tests, you clearly know what's going on or what could happen in unforeseen cases.

dragonmantank:
Just FYI, haven't forgotten about this just got pulled into another project for a few days that needed a bunch of attention!

SoftCreatR:
That's fine. I had the chance to take a short break 😂

SoftCreatR:
@dragonmantank Is it correct, that you tagged the latest version as 2.5.0? Wouldn't 2.4.2 be more suitable?

SoftCreatR:
FYI: This SDK isn't compatible with PHP 8, because of `laminas/laminas-diactoros`.

test/Voice/Webhook/FactoryTest.php
SoftCreatR: Done 😄

src/Voice/Filter/VoiceFilter.php
```diff
@@ -85,86 +95,114 @@ public function getQuery() : array
...
     }
 
-    public function setDateStart(\DateTimeImmutable $dateStart) : self
+    /**
+     * @return $this
+     */
+    public function setDateStart(DateTimeImmutable $dateStart): self
     {
-        $dateStart = $dateStart->setTimezone(new DateTimeZone('Z'));
+        $dateStart->setTimezone(new DateTimeZone('Z'));
```
dragonmantank, line 123: Need to check this, as the immutable class shouldn't be editable.

src/Voice/Filter/VoiceFilter.php
```diff
@@ -85,86 +95,114 @@ public function getQuery() : array
...
     }
 
-    public function setDateEnd(\DateTimeImmutable $dateEnd) : self
+    /**
+     * @return $this
+     */
+    public function setDateEnd(DateTimeImmutable $dateEnd): self
     {
-        $dateEnd = $dateEnd->setTimezone(new DateTimeZone('Z'));
+        $dateEnd->setTimezone(new DateTimeZone('Z'));
```
dragonmantank, line 139: Same with this

src/Voice/NCCO/Action/Connect.php
```diff
@@ -58,13 +66,14 @@ public function __construct(EndpointInterface $endpoint)
         $this->endpoint = $endpoint;
     }
 
-    public static function factory(EndpointInterface $endpoint, array $data = []) : Connect
+    public static function factory(EndpointInterface $endpoint): Connect
```
dragonmantank, line 69: Even though it's not used, can't remove it without it being a breaking change. Need to put back.

dragonmantank (changes requested):
Nearly done. Most are notes for me to check a few things, and I'm going to revert a bunch of test changes from `$this->` to `self::` to cut down on the size of the changeset. Both work and there is no official "correct" way so it will just shrink this down a bit. I'll handle that and push to the branch. 

Home stretch!

dragonmantank:
@SoftCreatR Can you open the PR to allow push access to the branch? Should be a checkbox on the PR that you can toggle. 

https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

SoftCreatR:
> @SoftCreatR Can you open the PR to allow push access to the branch? Should be a checkbox on the PR that you can toggle.
> 
> https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Done 😃

test/NetworkTest.php
```diff
@@ -8,26 +9,28 @@
...
 class NetworkTest extends TestCase
 {
-    public function testNetworkArrayAccess()
+    public function testNetworkArrayAccess(): void
     {
         $network = new Network('12345', 'Demo Network');
-        $this->assertEquals(@$network['network_code'], '12345');
-        $this->assertEquals(@$network['network_name'], 'Demo Network');
+
+        $this->assertEquals('12345', $network['network_code']);
```
dragonmantank, line 21: NOTE - Array access should throw a deprecation warning, need to check the underlying class

test/Numbers/ClientTest.php
```diff
@@ -316,40 +388,55 @@ public function testSearchOwnedPassesInAllowedAdditionalParameters()
...
             $this->assertEquals('rest.nexmo.com', $request->getUri()->getHost());
             $this->assertEquals('POST', $request->getMethod());
+
             return true;
         }))->willReturn($this->getResponse('post'));
 
-        $number = new Number('1415550100', 'US');
-        $this->numberClient->purchase($number);
+        //$number = new Number('1415550100', 'US');
+        @$this->numberClient->purchase('1415550100', 'US');
```
dragonmantank, line 429: Should support both ways of purchasing a number, so should have both tested.

test/Numbers/ClientTest.php
```diff
@@ -439,19 +578,25 @@ public function testCancelNumberWithNumberString()
...
+    public function testCancelNumberWithNumberAndCountryString(): void
     {
         // When providing a number string, the first thing that happens is a GET request to fetch number details
         $this->vonageClient->send(Argument::that(function (RequestInterface $request) {
             return $request->getUri()->getPath() === '/account/numbers';
         }))->willReturn($this->getResponse('single'));
 
-
-        // Then we get a POST request to cancel
+    // Then we get a POST request to cancel
```
dragonmantank, line 594: Weird indentation here, needs reverted

test/Redact/ClientTest.php
```diff
@@ -92,40 +119,83 @@ public function testOptionsDoNotOverwriteParams()
...
      */
-    public function testExceptions($response, $code, $expectedException, $expectedMessage)
+    public function testExceptions($response, $code, $expectedException, $expectedMessage): void
     {
         $this->expectException($expectedException);
         $this->expectExceptionMessage($expectedMessage);
 
         $this->vonageClient->send(Argument::that(function (RequestInterface $request) {
-            return true;
+            return $request instanceof RequestInterface;
```
dragonmantank, line 137: Same as others, should revert because this check is always true.

dragonmantank (changes requested):
Few more comments on things to revert and check, stopping here because Github is acting funny. 24 files left.

dragonmantank (changes requested):
The elephpant has been eaten.

Now going through all the notes I left and a final pass!

dragonmantank (changes requested):
This is provisionally approved! Running through a few more things before I hit the approve button

SoftCreatR:
Well, I'm a bit confused. What needs to be done now?

SoftCreatR:
Okaaaaaaaay.... 😄

dragonmantank:
> Well, I'm a bit confused. What needs to be done now?

Nothing! Just a bit of cleanup before the merge. I've got one more PR that will get added to master before this goes out as an official release, but THANK YOU very much for the work!

SoftCreatR:
> > Well, I'm a bit confused. What needs to be done now?
> 
> Nothing! Just a bit of cleanup before the merge. I've got one more PR that will get added to master before this goes out as an official release, but THANK YOU very much for the work!

Well, it was fun while it lasted. Let's see, if there's something for me to do in v3 :octocat:

PR #260, opened 2020-10-19
Changed ocramius/package-versions dependancy to composer/package-versions-deprecated to support Composer 2.0

footballencarta:
* Removed PHP 7.1 from travis, as package only supports 7.2 and above
* Added a couple of auto-generated files to .gitignore
* Changed dependancy of ocramius/package-versions to composer/package-versions-deprecated to allow for composer 2.0 to be supported (https://github.com/composer/package-versions-deprecated)


dragonmantank:
Thank you! This all looks good, but I'm going to hold off merging it until we get the 2.5.0 PR all sinched up as it also affects some of these same files. 

I've marked it as `hacktoberfest-accepted` so you should still get credit!

footballencarta:
Composer 2.0 was released yesterday (https://github.com/composer/composer/releases/tag/2.0.0) - so it might be worth looking at merging it sooner rather than later? As currently it doesn’t work with composer 2.0.

SoftCreatR:
@dragonmantank Just do it. If there are conflicts with 2.5.0, I'll resolve them.

dragonmantank:
This is merged, cutting a new release now

SoftCreatR:
I'll continue on 2.5 later this day.

PR #261, opened 2020-10-29
Request php error in setCodeLength logic

Norfaer:
Comparison should be AND for "not eqal" if you want to check value to be one of "4", "6"

### DIFF
--- a/src/Verify/Request.php
+++ b/src/Verify/Request.php
@@ -102,7 +102,7 @@ class Request implements ArrayHydrateInterface
 
     public function setCodeLength(int $codeLength) : self
     {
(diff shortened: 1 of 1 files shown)


dragonmantank:
Can we get some tests to show that the change works, as well as to avoid regressions? Thanks!

SoftCreatR:
https://github.com/Vonage/vonage-php-sdk-core/pull/257/commits/929e2b8051217cb45fbe9a00fdd4080757af026d

dragonmantank:
As mentioned this was ported back to what ended up becoming the basis for 2.6.0, but I added additional happy-path checks to make sure 4 and 6 worked. Thanks!

PR #270, opened 2020-12-07
Uses the Versions package to better determine the Guzzle version

dragonmantank:
## Description

This changes from using the class names to determine the guzzle version to using the actual version number reported in Composer. This allows us to determine with better accuracy which version of Guzzle is available so we can make the correct wrapper.

## Motivation and Context

Since we support both Guzzle 6 and Guzzle 7, and Guzzle 6 needs a wrapper, we need to have a good consistent way of checking this. 

## How Has This Been Tested?

Tested by forcing Guzzle 6 with the wrapper and 7 in various demo instances to make sure detection was working. Example composer setups are below. Each test successfully sent the SMS message using the `send-sms.php` code snippet code.

### Guzzle 6 Test
```console
$  composer require guzzlehttp/guzzle:^6.0 php-http/guzzle6-adapter vonage/client-core:dev-better-guzzle-detection@dev
$  php test.php
The message was sent successfully
```

### Guzzle 7 Test
```console
$  composer require guzzlehttp/guzzle:^7.0 vonage/client-core:dev-better-guzzle-detection@dev
$  php test.php
The message was sent successfully
```

## Types of changes

## Checklist:

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -124,14 +124,18 @@ class Client
         if (is_null($client)) {
             // Since the user did not pass a client, try and make a client
             // using the Guzzle 6 adapter or Guzzle 7 (depending on availability)
-            /** @noinspection ClassConstantCanBeUsedInspection */
-            if (class_exists('\GuzzleHttp\Client')) {
-                $client = new \GuzzleHttp\Client();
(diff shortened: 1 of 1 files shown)


GregHolmes (approved):
Can't see any problems with this.

PR #274, opened 2021-02-23
Added detail field to VAPI event object

dragonmantank:
## Description

Allows the `detail` field to be properly deserialized and accessed if it's part of an event.

## Motivation and Context

This field now appears on some events, so should be made available. Otherwise this will just return null.

## How Has This Been Tested?

Unit testing. I was not able to generate the new events using a US number so crafted a webhook response based on the API spec.

```php
// If the webhook is an Event with a detail key:
$event = Factory::createFromRequest($request);
echo $event->getDetail();
```

## Types of changes

## Checklist:

### DIFF
--- a/src/Voice/Webhook/Event.php
+++ b/src/Voice/Webhook/Event.php
@@ -41,6 +41,11 @@ class Event
     /**
      * @var string
      */
+    protected $detail;
+
+    /**
+     * @var string
+     */
     protected $direction;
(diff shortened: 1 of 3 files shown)


bencgreenberg (approved):

bencgreenberg:
Spent time this morning reviewing this, it looks good!

PR #275, opened 2021-02-25
Moved Secrets to own client

dragonmantank:
## Description

This moves all of the logic for Secrets management to it's own client and namespace, and deprecates all the old code.

## Motivation and Context

As part of a general move for v3.x and it's breaking changes, the Account client is pulling a lot of weight, as it is doing Secret Management, Pricing, and Account configuration (but not owned Numbers management, which is also under the Account API). While Secrets management resides under the Account API umbrella, it's an API that can live on it's own. 

## How Has This Been Tested?

Unit tests and test scripts.

Accessing the client can be done with:

```php
$client->secrets()->[method]();
```

## Types of changes

## Checklist:

### DIFF
--- a/src/Account/Client.php
+++ b/src/Account/Client.php
@@ -283,12 +283,15 @@ class Client implements ClientAwareInterface, APIClient
     }
 
     /**
+     * @deprecated use the Vonage\Secrets\Client::list method
+     *
      * @throws ClientExceptionInterface
      * @throws ClientException\Exception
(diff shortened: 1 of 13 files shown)


GregHolmes (approved):
Looks good to me, but you might want to update the snippets on the `php-code-snippets repository`

dragonmantank:
the current snippets won't break, but I'll get those updated here soon so they use the new format.

PR #285, opened 2021-04-19
Properly handle MT Throttling and SMS Throttle Time

### LINKED ISSUE #282: SMS Throttle Bug
What is the purpose of these lines?

https://github.com/Vonage/vonage-php-sdk-core/blob/7cfa39414ffb0529ee05d0f4f22225f43cc72ef6/src/SMS/ExceptionErrorHandler.php#L59-L61

In the below case, the script will sleep for 31 seconds and is causing SMS send to timeout (when script max execution time is set to let's say 30 seconds) as it picks the first digits, add 1 and sleep.

https://github.com/Vonage/vonage-php-sdk-core/blob/7cfa39414ffb0529ee05d0f4f22225f43cc72ef6/src/SMS/Client.php#L50-L54

However, sleeping for 1 second is always enough as the interval is always 1 second. Here's an example for exceeded max MT

```
{
    "message-count": "1",
    "messages": [
        {
            "status": "1",
            "error-text": "Exceeded maximum MT throughput[ MAX [ 30 ] INTERVAL [ 1000 ] COUNT [ 30 ] LAST INT-START [ 1617784922564 ] TIME_SINCE_LAST_INTERVAL_START [ 529 ]  ] "
        }
    ]
}
```

The MAX [30] is the MAX allowed MT throughput (API requests), if we want to sleep for the minimum amount of time we will have to use usleep and set the timeout (1000-529) in milli/microseconds.

dragonmantank:
## Description

Fixes #282 

This switches the exception handlers to deal with the proposed throttle time in milliseconds, with a minimum of 1 second. The code originally took the proposed value as seconds and increased it by that plus 1, leading to very long throttling times.

This fix also tightens the regex for determining throttle times so it only handles SMS throttles. Generic API message terminating limits now throttle for 1 second, since API throttles are based on messages per second.

## Motivation and Context

The SMS throttle messages can return very high proposed values as they are in milliseconds, not seconds. This leads to scripts timing out as values like "702" were interpreted as 702 seconds, not .702 seconds. We use a minimum of 1 as `sleep()` assumes seconds, and it's better to be slightly longer than too short and trigger another throttle.

The Regex fix stops the problem reported in #282 were we incorrectly grabbed a value because of a very loose regex.

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/src/Message/Client.php
+++ b/src/Message/Client.php
@@ -108,7 +108,7 @@ class Client implements ClientAwareInterface
                                 $e->setTimeout(1);
                                 $e->setEntity($data);
 
-                                if (preg_match('#\[\s+(\d+)\s+]#', $part['error-text'], $match)) {
+                                if (preg_match('#Throughput Rate Exceeded - please wait \[\s+(\d+)\s+] and retry#', $part['error-text'], $match)) {
                                     $seconds = max((int)$match[1] / 1000, 1);
                                     $e->setTimeout($seconds);
                                 }
--- a/src/Message/Client.php
+++ b/src/Message/Client.php
@@ -109,7 +109,8 @@ class Client implements ClientAwareInterface
                                 $e->setEntity($data);
 
                                 if (preg_match('#\[\s+(\d+)\s+]#', $part['error-text'], $match)) {
-                                    $e->setTimeout((int)$match[1] + 1);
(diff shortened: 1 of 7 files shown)


GregHolmes (approved):
All looks good to me!

PR #286, opened 2021-04-26
Explicitly set the sub if required

dragonmantank:
## Description

To work around a change in the JWT dependency we have, explicitly check for and set the `sub` parameter.

## Motivation and Context

In newer versions of `lcobucci/jwt` the `sub` header can no longer be set as an arbitrary claim. Unfortunately to keep the wide berth of PHP versions we use we have to keep 3.x compatibility and 4.x compatibility, so this fix helps bridge that gap some more.

## How Has This Been Tested?

Unit tests and local test scripts

## Types of changes

## Checklist:

### DIFF
--- a/test/Client/Credentials/KeypairTest.php
+++ b/test/Client/Credentials/KeypairTest.php
@@ -93,6 +93,39 @@ class KeypairTest extends TestCase
     }
 
     /**
+     * @link https://github.com/Vonage/vonage-php-sdk-core/issues/276
+     */
+    public function testExampleConversationJWTWorks()
+    {
+        $credentials = new Keypair($this->key, $this->application);
+        $claims = [
+            'exp' => strtotime(date('Y-m-d', strtotime('+24 Hours'))),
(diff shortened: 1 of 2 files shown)


test/Client/Credentials/KeypairTest.php
```diff
@@ -9,7 +9,12 @@
 
 namespace VonageTest\Client\Credentials;
 
+use InvalidArgumentException;
```
GregHolmes, line 12: I can't find where these exceptions are being used. Did you mean to add them here?

dragonmantank: Those must have just been left over from earlier tests. Removed.

GregHolmes (approved):

GregHolmes:
Probably overboard, but is it worth adding in a comment, to state when we're not covering older versions of PHP to remove it?

dragonmantank:
The old version will fail to load once we drop older PHP versions, so we should be good.

PR #300, opened 2021-12-09
Fix prophecy to use trait, update maintainer details

SecondeJK:
## Description
The test suite output has been clogged with a large amount of depreciation notices for the same issue for some time. This PR will solve this.

## Motivation and Context
PHPUnit will throw the following deprecation:
PHPUnit\Framework\TestCase::prophesize() is deprecated and will be removed in PHPUnit 10
In order to fix this, we have to move this method into a new trait. This trait requires a new package, which in turn means the minimum PHP version MUST be bumped up to 7.4. 7.4 has currently just entered LTS.

We are keeping 7.2 as a minimum requirement for production (although by v3 this will likely be bumped up) but adding the minimum PHP version to be 7.4 to allow PHPUnit to complete without deprecations.

Due to the nature of this being my second contribution, I have also modified the packagist details to list me as the sole new maintainer.

## How Has This Been Tested?
This PR updates the test suite itself, so its very aim is to make sure the test suite passes.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: [ '7.2', '7.3', '7.4', '8.0' ]
+        php: [ '7.4', '8.0' ]
     name: PHP ${{ matrix.php }} Test
 
     steps:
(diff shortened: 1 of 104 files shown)


superchilled (approved):
Looks good

PR #301, opened 2021-12-16
Removal of major deprecations

SecondeJK:
## Description
This PR aims to continue the process of simplifying the PHP Core library by stripping out major marked deprecations since the release of v.2.2

## Motivation and Context

* The depreciated `Calls` module has been removed, use the `Voice` module instead
* The depreciated `User` module has been removed
* Incomplete tests have been removed. These are difficult to navigate from a change of maintainer ownership, and are marked in such as way as to be addressed in the future either in a target to get CodeCov to 100%, or by the forthcoming changes to remove logic such as ArrayAccess for objects
* Skipped tests have been deleted, some are from 2016 and serve as code bloat without domain knowledge pointers
* PHP version support has been ADDED for 8.1. This has been done by adding the `ReturnTypeWillChange` core PHP annotation. This is only a sticking plaster, but is not concerning as they have been added to methods that are marked as deprecated anyway
* PHP version support has been DROPPED for 7.2, which is past EOL
* Tests have been modified to make sure JWTs can be decoded when using Keypair creds

## How Has This Been Tested?
The test suite has been changed to pass fully.

## Types of changes

## Checklist:


stof:
This is a PR that should target a 3.x branch, not 2.x.

SecondeJK:
@stof The strategy of how this library will be maintained has changed, since the team at Vonage have changed as well. As maintainers of this library has changed, we are to make some changes to make releases smaller and more often. As a result, the end goal for PRs such as this one will eventually be the roadmap @dragonmantank was on for v3, but to get there we don't want to do a huge v3 release with every deprecation removed. Rather remove them gradually, then release v3 where the v2 branch is in a state we want it.

stof:
Removing deprecations in a minor release goes against semver. Dropping semver seems like a bad idea to me, as it is the expectation of most devs using composer.

SecondeJK:
@stof This is 100% true re: semver, which is not going to be changed. My aim is to stop the branching in place for v2.X and v3.X so that we can get the library back to one main branch that uses tags for release - therefore this would be merged to the current 2.X branch, a new main branch is created and a release of v3.X can be tagged

SecondeJK:
main has been renamed. When this is merged, it will be released as 3.0.0. I'll be putting out a release note through the Vonage blog on the branching strategy

PR #330, opened 2022-09-23
Add Pay NCCO

SecondeJK:
This PR adds the new Pay NCCO

## Description
More information can be found in the NCCO documentation here:
https://developer.vonage.com/voice/voice-api/ncco-reference#pay

## Motivation and Context
Pay has been GA for sometime, and needs to be included in the SDK

## How Has This Been Tested?
Full unit tests have been added for the new NCCO type

## Types of changes

## Checklist:

### DIFF
--- a/src/Voice/NCCO/Action/Pay.php
+++ b/src/Voice/NCCO/Action/Pay.php
@@ -8,8 +8,6 @@ use InvalidArgumentException;
 
 class Pay implements ActionInterface
 {
-    protected const PROMPT_TYPES = ['CardNumber', 'ExpirationDate', 'SecurityCode'];
-
     protected const PERMITTED_VOICE_KEYS = ['language', 'style'];
 
     protected const PERMITTED_ERROR_KEYS = [
@@ -100,7 +98,7 @@ class Pay implements ActionInterface
             throw new InvalidArgumentException('type is required when setting a text prompt.');
         }
 
-        if (!in_array($prompts['type'], self::PROMPT_TYPES, true)) {
+        if (!array_key_exists($prompts['type'], self::PERMITTED_ERROR_KEYS)) {
             throw new InvalidArgumentException('invalid prompt type.');
         }
 
--- a/src/Voice/NCCO/Action/Pay.php
+++ b/src/Voice/NCCO/Action/Pay.php
@@ -1,0 +1,208 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Vonage\Voice\NCCO\Action;
+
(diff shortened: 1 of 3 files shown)


src/Voice/NCCO/Action/Pay.php
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * @return float
+     */
+    public function getAmount(): float
+    {
+        return $this->amount;
+    }
+
+    public function setAmount(float $amount): void
+    {
```
superchilled, line 63: Perhaps validate that amount is greater than `0`?

manchuck: It's not the best idea to have currency as a float. When converting to a different currency can lead to floating point errors. Best to use an int then use something like `bc_math` to adjust to the correct domination. 

JPY for example does not divide its currency.

src/Voice/NCCO/Action/Pay.php
```diff
@@ -0,0 +1,206 @@
...
+
+    public static function factory(array $data): Pay
+    {
+        $pay = new self();
+
+        if (array_key_exists('amount', $data)) {
+            $pay->setAmount($data['amount']);
+        } else {
+            throw new InvalidArgumentException('Amount is required for this action.');
+        }
```
manchuck, line 186: might it be better to put validation into its own method and just have the factory call it? Better for SOC and allows for others to validate the object without needing to create a new object.

manchuck (commented):
Looking good but I would suggest having currency as an int instead of float

test/Voice/NCCO/Action/PayTest.php
```diff
@@ -0,0 +1,204 @@
...
+        $payAction = Pay::factory([
+            'amount' => 9.99,
+            'currency' => 'eur'
+        ]);
+
+        $this->assertInstanceOf(Pay::class, $payAction);
+        $this->assertEquals('eur', $payAction->getCurrency());
+    }
+
+    public function testCannotAddCurrencyAsFloat(): void
```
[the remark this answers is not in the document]
SecondeJK: The reason I've included this is because of developers possibly getting currency and amount confused

src/Voice/NCCO/Action/Pay.php
SecondeJK: As briefly mentioned, the reason this is being stored as a float which is less than ideal is because the API endpoint actually takes a float when sending the body

manchuck (approved):

PR #335, opened 2022-10-14
Clarify that the `/check` endpoint is really the `check()` method

derickr:
## Description
At first I couldn't find what the `/check` end point meant, but it was really the check method which is already documented.
Also made it use PIN consistently, instead of using pin and code.

## Types of changes

- [X] Documentation

## Checklist:

- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.


SecondeJK (approved):
Love a good docs correction from reading from a different angle!

PR #341, opened 2022-11-16
security bump for laminas

SecondeJK:
This is a general security update to bump the Laminas request version.


SecondeJK:
Cannot merge to bump versions until all dependencies drop PHP7.4

PR #343, opened 2022-11-29
Feature/php 8 refactor 2

SecondeJK:
PHP 8 Refactor and deprecation drop.

## Description
This is a major PR to deprecate the EOL PHP7.4, while introducing new language features that come with PHP8. The automated refactors come from RectorPHP, while some of the original deprecations set for v3 have been pulled out.

## Motivation and Context
For security reasons, PHP7.4 is no longer supported by Vonage because it is EOL. Dropped 7.4 enables us to take advantage of optimisations such as constructor promotion. I've also started to introduce typed class properties.

As deprecations have been pulled out under the Numbers API client, it makes this a backwards-compatible breaking change, so has to be bumped as a major version.

I have also bumped the copyright headers.

## Types of changes

## Checklist:


test/Numbers/ClientTest.php
```diff
@@ -104,14 +113,14 @@ public function testUpdateNumber($payload, $id, $expectedId, $lookup): void
...
+            $number = $this->numberClient->update($payload, $id);
         } else {
-            $number = @$this->numberClient->update($payload);
+            $number = $this->numberClient->update($payload);
         }
 
         $this->assertInstanceOf(Number::class, $number);
         if ($payload instanceof Number) {
-            $this->assertSame($payload, $number);
+            $this->assertSame($payload->getId(), $number->getId());
```
SecondeJK, line 123: This behaviour isn't the same as before, and I'm not sure it's correct. We're not really checking how object mutate between the API call

dragonmantank: Yeah, with the new code it's better to check that the values are equal rather than checking the payload. We return a new object, so the payload will be a different object than what was passed in. I'd expand it to test each element for equalness

SMadani (approved):

rector.php
```diff
@@ -0,0 +1,22 @@
...
+use Rector\Set\ValueObject\LevelSetList;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->paths([
+        __DIR__ . '/src',
+        __DIR__ . '/test',
+    ]);
+
+    // register a single rule
+    $rectorConfig->rule(\Rector\Php74\Rector\Property\TypedPropertyRector::class);
```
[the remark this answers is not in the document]
SecondeJK: Yep, we're missing PHP7.4 features!

PR #351, opened 2023-01-09
GSM-7 warning handling for users

SecondeJK:
## Description
I've done several iterations around this behaviour, but after feedback from users and various issues - this PR should resolve what I believe to be a regular pain point for developers.

## Motivation and Context
GSM-7 is a specific character set that does not have the encoding overhead of having to convert to Unicode. As a result, sending an SMS via. GSM-7 will result is less multi-message delivery if the content is longer than a standard text.

Because of this, I have now defaulted a new SMS object to be `text`.

This developer has total control over this, however. If you want to change the encoding, use `setType('unicode')`

The SDK does not *force* the developer to use the correct type. It does, however, run the necessary checks via. a new regex to warn the user if they are likely to run into a potential problem, either because it's something that needs to be unicode but it set to text (resulting in the end device using incorrect characters to substitute) or sending something that could be sent as GSM7 as unicode (potentially more expensive, especially with big notification runs in an app.

## How Has This Been Tested?
Tests have been included to make sure that the appropriate warnings are thrown.

## Types of changes

## Checklist:


src/SMS/Message/SMS.php
```diff
@@ -13,27 +13,41 @@
...
-    public function __construct(string $to, string $from, protected string $message, string $type = 'unicode')
+    public function __construct(string $to, string $from, protected string $message, string $type = 'text')
     {
         parent::__construct($to, $from);
         $this->setType($type);
     }
 
+    public function encodingError(): bool
+    {
+        if ($this->getType() === 'unicode' && $this->isGsm7()) {
```
SMadani, line 35: You could simplify this function into a single expression (i.e. one `return`)

src/SMS/Message/SMS.php
```diff
@@ -13,27 +13,41 @@
...
+            return true;
+        }
+
+        if ($this->getType() === 'text' && ! $this->isGsm7()) {
+            return true;
```
SMadani, line 40: I would reconsider the `encodingError()` function altogether, since in this case, the error message does not reflect the cause, it covers the first case but not this. I'd set the error message on this class in this function, and if it's not null then throw it in `send`.

SMadani (changes requested):

src/SMS/Message/SMS.php
manchuck: Would it make sense to have encoding error be part of the message interface? You can then move the warning from client.php to here. For other message types, it can just be an empty function (Use a trait or have it on the OutboundMessage abstract class)

src/SMS/Message/SMS.php
SecondeJK: There two conditions for this - unicode set and the method returns true on GSM7... OR text set and the method returns false on GSM7. I decided to code this very explicitly, since the first time we came across unicode tests it was very much implicit knowledge. Doing this as one return would mean bundling 2 statements in a quadruple if statement which might be a bit confusing, so that's why I've done it this way

SecondeJK: I've changed the error reporting here through the inheritance structure

manchuck (approved):
It looks good. I thought that `$message->getErrorMessage()` would call `trigger_error`. But it's not a show stopper

PR #353, opened 2023-01-23
remove ext-json from composer.json

ankurk91:
* ext-json is always available from php 8.0+
https://php.watch/versions/8.0/ext-json

* no need to specify php version twice


SecondeJK:
This is a great spot, thanks!

SecondeJK (approved):

PR #362, opened 2023-02-02
Fix class name to identify client version

Froxz:
Class "Vonage\Versions" not found (v 4.0.3 and 4.0.2 are unusable currently)

```php
return Versions::getVersion('vonage/client-core');
```

Regression after [#354](https://github.com/Vonage/vonage-php-sdk-core/pull/354)


SecondeJK:
Great spot, caused by bad merge.

PR #382, opened 2023-03-03
Change warning to logger for GSM-7

SecondeJK:
We've been through several implementations of this, and I think we're at a final one for the behaviour of sending an SMS that can be used by frameworks effectively.

## Description
If you have chosen to set your SMS as unicode, it will no longer issue a warning if it contains a GSM-7 charset: you have explicitly told the code that you want to send it as unicode because it is not the default.

If you are sending a message that should be in unicode, but have not changed the type, a log warning will be fired *if you have set a logger up* for the library. We don't want to stop you sending the message - it might be garbled, but it's your choice to send it this way if you wish. To prevent this, make sure to use the `isGsm7` static method on the SMS object.

## Motivation and Context
Trying to get the best user experience we can (and it's hard)

## How Has This Been Tested?
One test deleted, the other changed to use prophecy to test the call to the logger.

## Types of changes

## Checklist:


test/SMS/ClientTest.php
```diff
@@ -377,57 +378,48 @@ public function testCanHandleMissingShortcodeOn2FA(): void
...
-        $this->expectWarning();
-        $this->expectErrorMessage("You are sending a message as `unicode` when it could be `text` or a
-            `text` type with unicode-only characters. This could result in increased billing - 
-            See https://developer.vonage.com/messaging/sms for details, or email support@vonage.com if you have any 
-            questions.");
+//        $this->expectWarning();
+//        $this->expectErrorMessage("You are sending a message as `text` when contains unicode only
+//            characters. This could result in encoding problems with the target device - See
+//            https://developer.vonage.com/messaging/sms for details, or email support@vonage.com if you have any
+//            questions.");
```
manchuck, line 391: Commented-out code makes MANCHUCK SMASH!!!

manchuck (changes requested):

SMadani (approved):

test/SMS/ClientTest.php
```diff
@@ -377,57 +378,48 @@ public function testCanHandleMissingShortcodeOn2FA(): void
...
         $this->vonageClient->send(Argument::that(function (Request $request) {
             return true;
         }))->willReturn($this->getResponse('send-success'));
 
-        $this->expectWarning();
-        $this->expectErrorMessage("You are sending a message as `unicode` when it could be `text` or a
-            `text` type with unicode-only characters. This could result in increased billing - 
-            See https://developer.vonage.com/messaging/sms for details, or email support@vonage.com if you have any 
-            questions.");
+//        $this->expectWarning();
```
dragonmantank: Delete the code rather than comment it out. Worst case we can retrieve it from git if we need it

test/SMS/ClientTest.php
SecondeJK: wow

manchuck (approved):

rector.php
```diff
@@ -1,22 +0,0 @@
-<?php
```
dragonmantank, line 1: Did we mean to nuke this file?

src/Client.php
```diff
@@ -72,7 +72,6 @@
  * Vonage API Client, allows access to the API from PHP.
  *
  * @method Account\Client account()
- * @method Message\Client message()
```
dragonmantank, line 75: Just a note, might get some people complain about this. I don't think there's a good way to mark it as deprecated though since it's a magic call.

src/Client.php
```diff
@@ -380,30 +380,8 @@ public function post(string $url, array $params): ResponseInterface
     }
 
     /**
-     * Takes a URL and a key=>value array to generate a POST PSR-7 request object
-     *
-     * @throws ClientExceptionInterface
-     * @throws ClientException
-     */
-    public function postUrlEncoded(string $url, array $params): ResponseInterface
```
dragonmantank, line 388: Needs to come back and marked as deprecated

src/SMS/Client.php
```diff
@@ -12,17 +12,20 @@
...
 
 use Psr\Http\Client\ClientExceptionInterface;
+use Psr\Log\LogLevel;
 use Vonage\Client\APIClient;
 use Vonage\Client\APIResource;
 use Vonage\Client\Exception\Exception as ClientException;
 use Vonage\Client\Exception\ThrottleException;
+use Vonage\Logger\LoggerTrait;
 use Vonage\SMS\Message\Message;
-use Vonage\SMS\Message\SMS;
```
dragonmantank, line 20: Was this just a bad import that was never originally removed?

src/SMS/Message/Message.php
```diff
@@ -15,5 +15,6 @@ interface Message
 {
     public function toArray(): array;
     public function getErrorMessage(): ?string;
-    public function setErrorMessage(?string $errorMessage): void;
```
dragonmantank, line 18: This needs to come back as well, and just wrap `setWarningMessage()`

src/SMS/Message/Vcard.php
```diff
@@ -38,4 +38,12 @@ public function getCard(): string
     {
         return $this->card;
     }
+
```
dragonmantank, line 41: For these child classes, can we just move `getErrorMessage()` to `OutboundMessage`?

src/SMS/Message/WAPPush.php
```diff
@@ -50,4 +50,12 @@ public function getValidity(): int
     {
         return $this->validity;
     }
+
```
dragonmantank, line 53: For these child classes, can we just move `getErrorMessage()` to `OutboundMessage`?

test/SMS/ClientTest.php
```diff
@@ -377,57 +378,42 @@ public function testCanHandleMissingShortcodeOn2FA(): void
...
-
-        $this->expectWarning();
-        $this->expectErrorMessage("You are sending a message as `text` when contains unicode only 
+        $logger->log(
+            Argument::that(function ($level) {
+                $this->assertEquals('warning', $level);
+                return true;
+            }),
+            Argument::that(function ($message) {
+                $this->assertEquals('You are sending a message as `text` when contains unicode only
```
dragonmantank, line 407:
```suggestion
                $this->assertEquals('You are sending a message as `text` when contains non-GSM7 
```

dragonmantank (changes requested):
Small text change with the error message to be more clear (since it's not really unicode vs ASCII but GSM7 characters)

Couple of places we can tighten up code, and a few other questions

src/Client.php
SecondeJK: It's already been totally removed from the SDK though. There's no method

rector.php
SecondeJK: Yep, it was a rector config file I accidentally pushed. Harmless, but no needed

test/SMS/ClientTest.php
SecondeJK: Already been nuked, shouldn't have been committed to comments

src/SMS/Client.php
SecondeJK: Yep

src/SMS/Message/Vcal.php
```diff
@@ -38,4 +38,12 @@ public function getEvent(): string
     {
         return $this->event;
     }
+
```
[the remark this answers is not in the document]
SecondeJK: yep that would be much nicer

dragonmantank (approved):
LGTM

PR #389, opened 2023-03-20
Verify v2

SecondeJK:
## Description
Verify v2 has been made available for some time to developers. This PR adds the functionality to the SDK to use this API.

## Motivation and Context
Evolution of Vonage products and codebases

## How Has This Been Tested?
Full test suite has been added for the new client.

## Types of changes

## Checklist:


.gitattributes
```diff
@@ -13,3 +13,4 @@
 # Ignore all differences in line endings for requests
 /test/SMS/requests/* text eol=crlf
 /test/Voice/requests/* text eol=crlf
+/test/Verify2/Fixtures/Requests/* text eol=crlf
```
manchuck, line 16: I would still recommend updating this to make all line endings crlf

src/Verify2/VerifyObjects/VerifyEvent.php
```diff
@@ -0,0 +1,127 @@
...
+        'type',
+        'channel',
+        'status',
+        'finalized_at',
+        'client_ref'
+    ];
+
+    public function __construct(array $data)
+    {
+        foreach (static::$mandatoryFields as $key) {
```
manchuck, line 29: If the signature changes, it could be difficult to read the event from the webhook. Maybe not throw the exception?

If not, I would suggest having it be additive and list out all the missing keys instead of throwing if one key is missing  

(WDDT) What does @dragonmantank think?

manchuck (changes requested):
I also got dinged for not allowing basic auth to verify v2. Is that something the PHP SDK allows already?

Aside from my other question about throwing the exception this is g2g

manchuck (approved):

.gitattributes
SecondeJK: I think I'm leaning into being more explicit here, in case a general catchall breaks something else and is harder to find what the problem is

src/Verify2/VerifyObjects/VerifyEvent.php
SecondeJK: The options were to have magic methods or have something more explicit. I wanted to go for explicit, but you're right. As soon as the signature changes, all the getters and setters have to change.

SecondeJK:
I'm confused about Auth: this Client uses Basic Auth first before trying a Keypair

Issue #387, opened 2023-03-10
401 Bad Credential While Downloading a Recording


nabinem:
I upgraded  sdk from 3.x to latest version 4.0.x and now downloading recording no longer works which use to work on 3.x 

## Current Behavior
I am getting 401 Bad Credential 
$data->getStatusCode(); is returning 401 instead of 200

## Steps to Reproduce (for bugs)
$basic  = new \Vonage\Client\Credentials\Basic('key', 'secret');
$keypair = new \Vonage\Client\Credentials\Keypair(
    file_get_contents(VONAGE_APPLICATION_PRIVATE_KEY_PATH),
    VONAGE_APPLICATION_ID
);

$client = new \Vonage\Client(new \Vonage\Client\Credentials\Container($basic, $keypair));
$data = $client->get($recordingUrl);
file_put_contents($recordingId.'.mp3', $data->getBody());

$data->getStatusCode(); is returning 401 instead of 200

## Your Environment

* Version used: 4.0.10
*  PHP 8.0 on nginx

SecondeJK:
I think I've finally spotted what the problem is here.

You're using the `get` on the Client directly rather than the Voice Client. The individual clients are configured with APIResource objects that have AuthHandler configurations. I'll make a note to get some nice pass-through methods, but for the time being you can pull out the `APIResource` which will handle the auth for you:

```
$data = $client->voice()->getApiResource()->get($recordingUrl);
```

SecondeJK:
Ticket put in Vonage workload to add `$client->voice()->getRecording($url)` syntax

PR #393, opened 2023-03-30
Add ability for Voice to get Recording

SecondeJK:
## Description
Some changes have needed to be made that are a little bit hacky in order to allow users to use the client to get a recording out from an NCCO response containing a download URL. This has become a problem since the Auth handling was introduced, meaning that using a base client to get a URL would not add the auth.

There is now the ability to get a recording from the following method:
`$client->voice()->getRecording($url);`

Under the hood, an optional parameter has been introduced to the ApiResource class get method.
By default, it always takes a json response and decodes it. As the mp3 is a StreamResponse, you can overide the decoding as an optional parameter, leaving it to the user to call `$response->getBody()->getContents();`

## Motivation and Context
Raised in #387 

## How Has This Been Tested?
A sample mp3 has been added and a test written.

## Types of changes

## Checklist:


maxkahan (approved):
LGTM

PR #396, opened 2023-04-13
Fix purchase number deprecation trigger_error

ganey:
The error is currently triggered if you pass a string, but warns you to pass a string.

Either the error is incorrect, or the error needs moving to only trigger when passing a Number object

## Description
I have moved the trigger_error to happen only when a Number object is passed in

## Motivation and Context
The error currently triggers if you pass a string which is wrong

## How Has This Been Tested?
With deprecations turned on in php.ini

## Types of changes

## Checklist:

### DIFF
--- a/src/Numbers/Client.php
+++ b/src/Numbers/Client.php
@@ -225,14 +225,13 @@ class Client implements APIClient
                     "You must supply a country in addition to a number to purchase a number"
                 );
             }
-
+            $number = new Number($number, $country);
+        } else {
             trigger_error(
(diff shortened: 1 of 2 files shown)


SecondeJK:
This is a useful patch to make the behaviour make more sense - can you amend the tests to fix them with the new behaviour?

ganey:
Sorry i forgot to update the tests before. I have updated the testPurchaseNumberWithNumberObject to handle the deprecation

SecondeJK:
Apologies, there was slightly more work to do to fix this flow in the client so I have pushed in a change that partially implements this (but also adds the original intended functionality of passing in strings for the number and country)

PR #402, opened 2023-04-27
Feature/proactive connect

SecondeJK:
Proactive Connect draft PR

## Description
This PR adds Proactive Connect to the SDK

## Motivation and Context
Evolution of Vonage API product offerings

## How Has This Been Tested?
Full test suite added. No tests added for standard API response errors, as these are tested in other new SDKs AND in the `ApiResource` object

## Types of changes

## Checklist:


src/ProactiveConnect/Objects/ListBaseObject.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Vonage\ProactiveConnect\Objects;
+
+abstract class ListBaseObject
```
Tr00d, line 5: Is there a use for this abstract class?

EDIT: Nevermind, I found it. Personal opinion: I would have done a `marker interface` instead, given you don't have to share any code/behaviour.

Tr00d (approved):
LGTM

src/ProactiveConnect/Objects/ListBaseObject.php
SecondeJK: It's quite common to use abstract classes like this, which is specifically for typehinting. You can use an interface for this, but I want the typehint to be something that is extended

PR #414, opened 2023-06-14
Subaccounts API

SecondeJK:
This feature adds the subaccounts API to the base Client.

## Description
See documentation changes in the README for usage.
You can now access the subaccounts API using `$client->subaccount()`

## Motivation and Context
Evolution of Vonage API/SDKs

## How Has This Been Tested?
Full unit test suite has been added. Run it:
`./vendor/bin/phpunit --testsuite subaccount`

Note: does not contain error handling from the API, as this is tested within the test class for `ApiResource`.

## Types of changes

## Checklist:


src/Subaccount/Filter/Subaccount.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace Vonage\Subaccount\Filter;
+
+use Vonage\Client\Exception\Request;
+use Vonage\Entity\Filter\FilterInterface;
+
+class Subaccount implements FilterInterface
+{
+    public ?string $startDate = null;
```
SMadani, line 10: Start date is required. Either set the default here or in the client. Default can be the epoch.

src/Subaccount/Request/TransferRequest.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->to = $data['to'];
+        $this->amount = $data['amount'];
+        $this->reference = $data['reference'];
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        // TODO harden me with array key checks pls
```
SMadani, line 90: Do TODO? Also reference is optional so if it's null it should be omitted.

src/Subaccount/Request/TransferRequest.php
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Vonage\Subaccount\Request;
+
+use Vonage\Entity\Hydrator\ArrayHydrateInterface;
+
+class TransferRequest implements ArrayHydrateInterface
+{
+    private string $from;
+    private string $to;
+    private string $amount;
```
SMadani, line 11: Should be a decimal number.

src/Subaccount/SubaccountObjects/BalanceTransfer.php
```diff
@@ -0,0 +1,130 @@
...
+            $this->setCreatedAt($data['created_at']);
+        }
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        return [
+            'balance_transfer_id' => $this->getBalanceTransferId(),
```
SMadani, line 122: `balance_transfer_id` should be `id` (the spec is wrong).

src/Subaccount/SubaccountObjects/CreditTransfer.php
```diff
@@ -0,0 +1,124 @@
...
+            $this->setCreatedAt($data['created_at']);
+        }
+
+        return $this;
+    }
+
+    public function toArray(): array
+    {
+        return [
+            'credit_transfer_id' => $this->getCreditTransferId(),
```
SMadani, line 116: Don't know if the spec is correct since for balance transfers it's just `id` so in theory this is the same object as BalanceTransfer.

test/Subaccount/SubaccountObjects/AccountTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->assertEquals('John Doe', $instance->getName());
+
+        $instance->setPrimaryAccountApiKey('def456');
+        $this->assertEquals('def456', $instance->getPrimaryAccountApiKey());
+
+        $instance->setUsePrimaryAccountBalance(true);
+        $this->assertTrue($instance->getUsePrimaryAccountBalance());
+
+        $instance->setCreatedAt('2023-06-06');
+        $this->assertEquals('2023-06-06', $instance->getCreatedAt());
```
SMadani, line 53: Note that the format should be something like `2023-06-26T10:23:59Z`.

SMadani (changes requested):

src/Subaccount/Request/NumberTransferRequest.php
```diff
@@ -0,0 +1,102 @@
...
+
+class NumberTransferRequest implements ArrayHydrateInterface
+{
+    private string $from;
+    private string $to;
+    private string $number;
+    private string $country;
+    protected string $apiKey;
+
+    public function __construct(string $apiKey)
```
dragonmantank, line 15: Are `from`, `to`, `number`, and `country` not required to do the transfer? If they are all required they should come through the constructor to force the user to supply the information. Allowing a builder pattern here gives the user an opportunity to forget something.

src/Subaccount/Request/TransferRequest.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
+namespace Vonage\Subaccount\Request;
+
+use Vonage\Entity\Hydrator\ArrayHydrateInterface;
+
+class TransferRequest implements ArrayHydrateInterface
```
dragonmantank, line 7: Should rename this to `TransferCreditRequest` to be clearer

src/Subaccount/Request/TransferRequest.php
dragonmantank: For numerical safety because this is money, the user should pass in a string. We can cast it to a decimal later, but otherwise we're dealing with float issues.

src/Subaccount/Request/TransferRequest.php
```diff
@@ -0,0 +1,98 @@
...
+{
+    private string $from;
+    private string $to;
+    private string $amount;
+    private string $reference;
+
+    public function __construct(protected string $apiKey)
```
dragonmantank, line 14: If `to`, `from`, and `amount` are required (which makes sense) they should be moved to the constructor, not as builder options.

test/Subaccount/ClientTest.php
```diff
@@ -0,0 +1,369 @@
...
+                mb_substr($request->getHeaders()['Authorization'][0], 0, 6)
+            );
+            return true;
+        }))->willReturn($this->getResponse('get-success'));
+
+        $apiKey = 'something';
+        $response = $this->subaccountClient->getPrimaryAccount($apiKey);
+    }
+
+    public function testWillPatchSubaccount(): void
```
dragonmantank, line 68: I know it's using a PATCH method, but from a "what are we doing" perspective we are updating the subaccount

```suggestion
    public function testWillUpdateSubaccount(): void
```

test/Subaccount/ClientTest.php
```diff
@@ -0,0 +1,369 @@
...
+    {
+        $apiKey = 'acc6111f';
+        $subaccountKey = 'bbe6222f';
+
+        $payload = [
```
dragonmantank, line 73: Why are we using a raw array when we have an `Account` DTO?

test/Subaccount/ClientTest.php
```diff
@@ -0,0 +1,369 @@
...
+
+        $response = $this->subaccountClient->getPrimaryAccount($apiKey);
+        $this->assertInstanceOf(Account::class, $response);
+    }
+
+    public function testWillCreateSubaccount(): void
+    {
+        $apiKey = 'acc6111f';
+
+        $payload = [
```
dragonmantank, line 120: Why are we using a raw array instead of the `Account` DTO?

src/Subaccount/Client.php
```diff
@@ -0,0 +1,121 @@
...
+
+        $hydrator = new AccountHydrator();
+        $subaccounts = $collection->getPageData()['_embedded'][$api->getCollectionName()];
+
+        return array_map(function ($item) use ($hydrator) {
+            return $hydrator->hydrate($item);
+        }, $subaccounts);
+    }
+
+    public function createSubaccount(string $apiKey, array $payload): ?array
```
dragonmantank, line 62: We should be using DTOs to force the user to submit the correct data. Arrays make it too easy for the user to forget something, or forces them to have the documentation handy.

src/Subaccount/Client.php
```diff
@@ -0,0 +1,121 @@
...
+        return (new BalanceTransfer())->fromArray($response);
+    }
+
+    public function makeCreditTransfer(TransferRequest $transferRequest): CreditTransfer
+    {
+        $response = $this->api->create($transferRequest->toArray(), '/' . $transferRequest->getApiKey() . '/credit-transfers');
+        return (new CreditTransfer())->fromArray($response);
+    }
+
+    public function updateSubaccount(string $apiKey, string $subaccountApiKey, array $update): ?array
```
dragonmantank, line 80: We should be using DTOs to force the user to submit the correct data. Arrays make it too easy for the user to forget something, or forces them to have the documentation handy.

src/Subaccount/ClientFactory.php
```diff
@@ -0,0 +1,23 @@
...
+use Vonage\Verify2\Client;
+
+class ClientFactory
+{
+    public function __invoke(ContainerInterface $container): Client
+    {
+        $api = $container->make(APIResource::class);
+        $api->setIsHAL(true)
+            ->setErrorsOn200(false)
+            ->setClient($this->vonageClient->reveal())
```
dragonmantank, line 18: Is this a copy/paste error? This looks like a test mock but this isn't a test class.

src/Subaccount/Hydrators/AccountHydrator.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
dragonmantank, line 1: Why aren't you using the stock `ArrayHydrator`? You only need to have custom hydrators if are doing something special.

src/Subaccount/Hydrators/BalanceTransferHydrator.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Vonage\Subaccount\Hydrators;
+
+use Vonage\Entity\Hydrator\HydratorInterface;
+use Vonage\Subaccount\SubaccountObjects\Account;
+use Vonage\Subaccount\SubaccountObjects\BalanceTransfer;
+use Vonage\Subaccount\SubaccountObjects\CreditTransfer;
+
+class BalanceTransferHydrator implements HydratorInterface
```
dragonmantank, line 12: Why aren't you using the stock `ArrayHydrator`? You only need to have custom hydrators if are doing something special.

src/Subaccount/Hydrators/CreditTransferHydrator.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
dragonmantank, line 1: Why aren't you using the stock `ArrayHydrator`? You only need to have custom hydrators if are doing something special.

test/Subaccount/ClientTest.php
```diff
@@ -0,0 +1,369 @@
...
+        foreach ($response as $item) {
+            $this->assertInstanceOf(Account::class, $item);
+        }
+    }
+
+    public function testWillTransferCredit(): void
+    {
+        $apiKey = 'acc6111f';
+
+        $payload = [
```
dragonmantank, line 197: Marked the client method, but these things should be DTOs, not raw arrays. Alternatively, we should be requiring these values directly instead of hiding them in an array.

src/Subaccount/Filter/Subaccount.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace Vonage\Subaccount\Filter;
+
+use Vonage\Client\Exception\Request;
+use Vonage\Entity\Filter\FilterInterface;
+
+class Subaccount implements FilterInterface
```
dragonmantank, line 8: This is namespaced properly, but should make this Hungarian since we do that in most places.
```suggestion
class SubaccountFilter implements FilterInterface
```

dragonmantank (changes requested):

src/Subaccount/Request/TransferRequest.php
SecondeJK: This was a tentative TODO which is why I left it in there. There's no need for them though, I'll ditch them

src/Subaccount/Request/TransferRequest.php
SecondeJK: I actually missed the cast off, so that's a good spot. As tank says though, PHP can get awfully rough if you set it to a float initially

src/Subaccount/SubaccountObjects/CreditTransfer.php
SecondeJK: I'd quite like to keep these objects separate as it's clear to the user, despite the fact that the API has muddied the waters when treating the two types of transaction the same

src/Subaccount/ClientFactory.php
SecondeJK: aaah man

SMadani (approved):

PR #415, opened 2023-06-14
Output typo for backend when using AMD

SecondeJK:
This is a small patch PR to change the output of Advanced Machine Detection to the US spelling of behaviour.


manchuck (approved):
I mistype it like this all the time

PR #420, opened 2023-07-24
Add Users Client

SecondeJK:
This PR adds the ability to CRUD users with a new Users client.

## Description
These endpoints have been moved from the Conversations API, into the Application API. It seems most logical to move them under their own Client namespace.

## How Has This Been Tested?
Tests have been created for this new client.

## Types of changes

## Checklist:


src/Application/Client.php
```diff
@@ -18,10 +18,12 @@
 use Vonage\Client\ClientAwareInterface;
 use Vonage\Client\ClientAwareTrait;
 use Vonage\Client\Exception\Exception as ClientException;
+use Vonage\Entity\Filter\EmptyFilter;
+use Vonage\Entity\Filter\KeyValueFilter;
 use Vonage\Entity\Hydrator\ArrayHydrator;
 use Vonage\Entity\Hydrator\HydratorInterface;
 use Vonage\Entity\IterableAPICollection;
-
+use Vonage\Users\User;
```
dragonmantank, line 26: Can probably remove this since it's moved to it's own client

src/Application/Client.php
```diff
@@ -18,10 +18,12 @@
 use Vonage\Client\ClientAwareInterface;
 use Vonage\Client\ClientAwareTrait;
 use Vonage\Client\Exception\Exception as ClientException;
+use Vonage\Entity\Filter\EmptyFilter;
```
dragonmantank, line 21: Leftovers from before moving the users search to it's own client?

src/Application/Client.php
```diff
@@ -18,10 +18,12 @@
...
+use Vonage\Entity\Filter\KeyValueFilter;
```
dragonmantank, line 22: Leftovers from before moving the users search to it's own client?

phpunit.xml.dist
```diff
@@ -41,6 +41,9 @@
     <testsuite name="meetings">
       <directory>test/Meetings</directory>
     </testsuite>
+    <testsuite name="application">
+      <directory>test/Application</directory>
+    </testsuite>
```
dragonmantank, line 46: Should also add one for Users

src/Users/Client.php
```diff
@@ -0,0 +1,109 @@
...
+{
+    use ClientAwareTrait;
+
+    public function __construct(protected APIResource $api, protected ?HydratorInterface $hydrator = null)
+    {
+    }
+
+    public function getApiResource(): APIResource
+    {
+        if (is_null($this->api)) {
```
dragonmantank, line 37: `$this->api` is required, so we don't need these checks in new clients. That's a holdover from v2 and v3 where we allowed them to either be injected (the proper way) or built them for the user (in case they were using the clients directly)

src/Users/Client.php
```diff
@@ -0,0 +1,109 @@
...
+            $api->setClient($this->getClient())
+                ->setBaseUri('/v1/users')
+                ->setCollectionName('users');
+            $this->api = $api;
+        }
+
+        return $this->api;
+    }
+
+    public function getUsers($pageSize = null, $order = null, $cursor = null): IterableAPICollection
```
dragonmantank, line 48: Easier way is to create a custom filter for this with the page size, order, and cursor options. , then all you need to do is set it to `null` by default, and have then just make an `EmptyFilter` if the user doesn't pass anything in.

Also the name should be `listUsers` by convention.

src/Users/Client.php
```diff
@@ -0,0 +1,109 @@
...
+        $response->setPageSizeKey('page_size');
+        $response->setHasPagination(false);
+
+        return $response;
+    }
+
+    public function createUser(User $user): User
+    {
+        $response = $this->api->create($user->toArray());
+        $userObject = new User();
```
dragonmantank, line 76: You can cut down on the copy/paste by passing in a hydrator that is pre-set to the `User` class. Then you can just hydrate from there.

src/Users/ClientFactory.php
```diff
@@ -0,0 +1,33 @@
...
+class ClientFactory
+{
+    public function __invoke(ContainerInterface $container): Client
+    {
+        /** @var APIResource $api */
+        $api = $container->make(APIResource::class);
+        $api
+            ->setBaseUri('/v1/users')
+            ->setCollectionName('users')
+            ->setAuthHandler(new BasicHandler());
```
dragonmantank, line 29: Users API uses a JWT, not Basic

src/Users/User.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Vonage\Users;
+
+use Vonage\Entity\Hydrator\ArrayHydrateInterface;
+
+class User implements ArrayHydrateInterface
+{
+    private array $data;
+
+    public function __get($property)
```
dragonmantank, line 11: Don't use magic methods as it makes it extremely hard for IDEs to automatically show intelesense info. While you could supplement this with an annotation, since we allow users to also use this object to make Users we should enforce proper getters/setters.

src/Verify2/VerifyObjects/VerifyEvent.php
```diff
@@ -2,7 +2,9 @@
 
 namespace Vonage\Verify2\VerifyObjects;
 
-class VerifyEvent
+use Vonage\Entity\Hydrator\ArrayHydrateInterface;
```
dragonmantank, line 5: This should be split into a separate PR, this PR is about adding Users, not modifying Verify

src/Verify2/VerifyObjects/VerifyStatusUpdate.php
```diff
@@ -2,7 +2,9 @@
 
 namespace Vonage\Verify2\VerifyObjects;
 
-class VerifyStatusUpdate
+use Vonage\Entity\Hydrator\ArrayHydrateInterface;
```
dragonmantank, line 5: This should be split into a separate PR, this PR is about adding Users, not modifying Verify

test/Application/ClientTest.php
```diff
@@ -14,23 +14,24 @@
...
 use Vonage\Application\MessagesConfig;
 use Vonage\Application\RtcConfig;
 use Vonage\Application\VoiceConfig;
+use Vonage\Application\Webhook;
+use Vonage\Application\Webhook as ApplicationWebhook;
 use Vonage\Client;
 use Vonage\Client\APIResource;
 use Vonage\Client\Exception\Exception as ClientException;
 use Vonage\Client\Exception\Server as ServerException;
+use Vonage\Users\User;
```
dragonmantank, line 32: Probably a leftover form before this was moved to it's own class?

test/Users/ClientTest.php
```diff
@@ -0,0 +1,255 @@
...
+    protected APIResource $apiClient;
+
+    protected UsersClient $usersClient;
+
+    public function setUp(): void
+    {
+        $this->vonageClient = $this->prophesize(Client::class);
+        $this->vonageClient->getApiUrl()->willReturn('https://api.nexmo.com');
+        $this->vonageClient->getCredentials()->willReturn(
+            new Client\Credentials\Container(new Client\Credentials\Basic('abc', 'def'))
```
dragonmantank, line 42: Should be using JWT auth

src/Users/Client.php
```diff
@@ -0,0 +1,109 @@
...
+        $userObject->fromArray($response);
+
+        return $userObject;
+    }
+
+    public function getUserById(string $id): User
```
dragonmantank, line 82: This should just be `getUser()`, since it takes the ID. If this allowed other search params it could be `getUserBy[X]()`, but since we are passing the ID it should just be `getUser()`

src/Users/Client.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        $response = $this->api->get($id);
+        $returnUser = new User();
+        $returnUser->fromArray($response);
+
+        return $returnUser;
+    }
+
+    public function updateUser(User $user, string $id): User
```
dragonmantank, line 91: Should not need to pass an `$id` for this, because the ID is part of the `User` object. Instead, we should throw an exception if `$user->id` is not set.

src/Users/Client.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        $response = $this->api->partiallyUpdate($id, $user->toArray());
+        $returnUser = new User();
+        $returnUser->fromArray($response);
+
+        return $returnUser;
+    }
+
+    public function deleteUserById(string $id): bool
```
dragonmantank, line 100: Same as `getUserById()`, this should just be `deleteUser()` since it takes the ID.

test/Users/responses/delete-user-success.json
```diff
@@ -0,0 +1 @@
+{}
```
dragonmantank, line 1: This shouldn't exist as there is no response for deleting a user, just an empty body (`null`) and a 204 response

test/Users/ClientTest.php
```diff
@@ -0,0 +1,255 @@
...
+            $uriString = $uri->__toString();
+            $this->assertEquals(
+                'https://api.nexmo.com/v1/users/USR-82e028d9-5201-4f1e-8188-604b2d3471ec',
+                $uriString
+            );
+
+            $this->assertEquals('DELETE', $request->getMethod());
+
+            return true;
+        }))->willReturn($this->getResponse('delete-user-success', 204));
```
dragonmantank, line 240: This should return a new Response with a `null` body and a 204, not a string body of `{}`

dragonmantank (changes requested):

src/Users/Client.php
```diff
@@ -0,0 +1,88 @@
...
+    public function __construct(protected APIResource $api, protected ?HydratorInterface $hydrator = null)
+    {
+    }
+
+    public function getApiResource(): APIResource
+    {
+        return $this->api;
+    }
+
+    public function listUsers($pageSize = null, $order = null, $cursor = null): IterableAPICollection
```
dragonmantank, line 32:
```suggestion
    public function listUsers(FilterInterface $filter = null): IterableAPICollection
```

src/Users/Client.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        if (is_null($pageSize) && is_null($order) && is_null($cursor)) {
+            $filter = new EmptyFilter();
+        } else {
+            $filter = new UserFilter();
+            $filter->setPageSize($pageSize);
+            $filter->setOrder($order);
+            $filter->setCursor($cursor);
+        }
```
dragonmantank, line 41:
```suggestion
        if (is_null($filter)) {
            $filter = new EmptyFilter();
        } 
```

src/Users/Client.php
```diff
@@ -0,0 +1,88 @@
...
+namespace Vonage\Users;
+
+use Vonage\Client\APIClient;
+use Vonage\Client\APIResource;
+use Vonage\Client\ClientAwareInterface;
+use Vonage\Client\ClientAwareTrait;
+use Vonage\Client\Exception\Exception as ClientException;
+use Vonage\Entity\Filter\EmptyFilter;
+use Vonage\Entity\Hydrator\HydratorInterface;
+use Vonage\Entity\IterableAPICollection;
```
dragonmantank, line 14:
```suggestion
use Vonage\Entity\IterableAPICollection;
use Vonage\Entity\Filter\FilterInterface;
```

src/Users/Client.php
dragonmantank: Just have the user pass in a filter directly

src/Users/Client.php
dragonmantank: Then we can just make an empty filter if they don't pass anything at all

src/Users/Hydrator.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
dragonmantank, line 1: We can delete this and just use the `ArrayHydrator`, all the hydrators take a "prototype" class to know how what object to hydrate

src/Users/ClientFactory.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Vonage\Users;
+
+use Psr\Container\ContainerInterface;
+use Vonage\Client\APIResource;
+use Vonage\Client\Credentials\Handler\KeypairHandler;
```
dragonmantank, line 9:
```suggestion
use Vonage\Client\Credentials\Handler\KeypairHandler;
use Vonage\Entity\Hydrator\ArrayHydrator;
```

src/Users/ClientFactory.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    public function __invoke(ContainerInterface $container): Client
+    {
+        $api = $container->make(APIResource::class);
+        $api
+            ->setBaseUri('/v1/users')
+            ->setCollectionName('users')
+            ->setAuthHandler(new KeypairHandler());
+
+        return new Client($api, new Hydrator());
```
dragonmantank, line 21:
```suggestion
        $hydrator = (new ArrayHydrator())->setPrototype(new User());
        return new Client($api, $hydrator);
```

src/Users/Client.php
```diff
@@ -0,0 +1,88 @@
...
+        $response = $this->api->create($user->toArray());
+
+        return $this->hydrator->hydrate($response);
+    }
+
+    public function getUser(string $id): User
+    {
+        $response = $this->api->get($id);
+        $returnUser = new User();
+        $returnUser->fromArray($response);
```
dragonmantank, line 63:
```suggestion
        return $this->hydrator->hydrate($response);
```

src/Users/ClientFactory.php
dragonmantank: `ArrayHydrator` takes a class to use as a sample, so we can tell the generic hydrator how to work without needing a custom one.

dragonmantank (changes requested):
Just a couple of other ways to cut down on code

src/Users/Filter/UserFilter.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public function getPageSize(): int
+    {
+        return $this->pageSize;
+    }
+
+    /**
+     * @return $this
+     */
```
manchuck, line 53: Not needed. IDE's will pick this up

src/Users/Client.php
SecondeJK: did all the work for the filter, didn't do the last bit!

manchuck (changes requested):

src/Users/Filter/UserFilter.php
SecondeJK: must have missed this. I tend to not use docblocks anymore unless unusual behaviour needs to be documented, because typehinting and return types handle all this

src/Users/User.php
```diff
@@ -0,0 +1,163 @@
...
+            $this->setImageUrl($data['image_url']);
+        }
+
+        if (isset($data['properties'])) {
+            $this->setProperties($data['properties']);
+        }
+
+        if (isset($data['channels'])) {
+            $this->setChannels($data['channels']);
+        }
```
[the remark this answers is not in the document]
SecondeJK: More unreadable for me tbh
I get what it's doing, but feels more magic

manchuck (approved):

PR #422, opened 2023-08-01
Code Example not working - $numbers is not an Iterateable Object

eimers:
## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,7 +413,7 @@ All `$client->voice()` methods require the client to be constructed with a `Vona
 `Vonage\Client\Credentials\Container` that includes the `Keypair` credentials:
 
 ```php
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -753,8 +753,9 @@ filter to narrow down your search.
 To purchase a number, you can pass in a value returned from number search:
 
 ```php
+/** @var Vonage\Numbers\Number[] $numbers */
 $numbers = $client->numbers()->searchAvailable('US');
-$number = $numbers->current();
+$number = array_shift($numbers);
```
SecondeJK, line 758:
```suggestion
$number = $numbers[0];
```

SecondeJK (changes requested):
I'd prefer it if we picked out the first number from the array, rather than shifting it as I think it's a bit of an ambiguous way to deal with pointers

PR #444, opened 2023-10-13
lcobucci/jwt v5

erickskrauch:
## Description
After merging https://github.com/Vonage/vonage-php-sdk-core/pull/441 there is no reason not to allow v5 of `lcobucci/jwt`.

## Motivation and Context
I still need v5 of that library on my project :)

## How Has This Been Tested?
Yes.

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     "php": "~8.0 || ~8.1 || ~8.2", 
     "ext-mbstring": "*",
     "laminas/laminas-diactoros": "^3.0",
-    "lcobucci/jwt": "^3.4|^4.0",
+    "lcobucci/jwt": "^4.0|^5.0",
     "psr/container": "^1.0 | ^2.0",
     "psr/http-client-implementation": "^1.0",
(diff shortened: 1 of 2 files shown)


SecondeJK:
In order to bump jwt to 5.0, it needs to happen here as this is the new dependency:
https://github.com/Vonage/vonage-php-jwt

I can open it there soon. Regarding JWT being on this project, that's only for backwards compatibility and will be removed as a dependency in the next major version (5)

erickskrauch:
@SecondeJK, you can take a look on [my PR](https://github.com/Vonage/vonage-php-jwt/pull/5) for your JWT wrapper library :) Anyway, I think this PR must be merged because there is no available beta for a major update of this library, which can be offered to your customers (which I'm). So it will happen somewhere in the future with no defined ETA. You just limit your customers with legacy dependencies with no other option. My PR removes that limitation and allows us to use both versions of the library until the older version is dropped in the next major update.

So once again: there is no alternative at the current moment and thus this PR needs to be applied.

PR #446, opened 2023-10-18
Fix async behavior of the NCCO record action

erickskrauch:
## Description
As described in the [documentation](https://developer.vonage.com/en/voice/voice-api/ncco-reference#record), the `record` action is asynchronous by default, unless it has `endOnSilence`, `timeOut` or `endOnKey` params. The current implementation of the `Record` action in this library provides no way to not specify that param.

## Motivation and Context
We need to achieve an async behavior of the record action.

## How Has This Been Tested?
We have made calls with those changes.

Call `CON-b4cc19a0-ffd7-4e44-80ec-2d7ecf05c44c`.

## Types of changes

## Checklist:

### DIFF
--- a/src/Voice/NCCO/Action/Record.php
+++ b/src/Voice/NCCO/Action/Record.php
@@ -53,7 +53,7 @@ class Record implements ActionInterface
     /**
      * @var int
      */
-    protected $timeOut = 7200;
+    protected $timeOut = 0;
 
     /**
      * @var bool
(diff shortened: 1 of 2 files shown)


SecondeJK:
This can be merged when it is time to release 5.0.0 (as although it's a fix, it's a semver breaking change in line with how the Record NCCO works currently)

SecondeJK (approved):

SecondeJK:
I've just picked up an internal ticket and fixed this in a different PR, although the solution seems to be much the same. Thanks for opening it though! #493

PR #465, opened 2024-01-18
New Messages API Features

SecondeJK:
This PR provides parity with new Messages API features.

## Description
`webhook_url` and `webhook_version` have been added to all Messages API Objects
`sms` property added to outbound SMS text objects, containing `encoding_type`, `content_id`, `entity_id` and `ttl`
MMS vCard now has `caption`
WhatsApp File object now has `name`

Fixed the Viber Video object which was missing the `viber_service` object, but now has mandatory `duration` and `file_size` keys

## Motivation and Context
Parity with Vonage APIs

## How Has This Been Tested?
Existing tests have been modified

## Types of changes

## Checklist:

### DIFF
--- a/src/Messages/Channel/BaseMessage.php
+++ b/src/Messages/Channel/BaseMessage.php
@@ -9,6 +9,13 @@ abstract class BaseMessage implements Message
     protected string $from;
     protected string $channel;
     protected ?string $clientRef = null;
+    protected ?string $webhookUrl = null;
+    protected ?string $webhookVersion = null;
+
+    protected array $permittedVersions = [
+        'v0.1',
+        'v1'
+    ];
 
     public const MESSAGES_SUBTYPE_TEXT = 'text';
     public const MESSAGES_SUBTYPE_IMAGE = 'image';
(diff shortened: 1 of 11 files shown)


superchilled (commented):
Looks good overall. I just had a question about `Messages/MessageObjects`. Are the same objects (and therefore properties) used across different channels? For example, with `FileObject`, the `__construct` function can set properties on the object for `url`, `caption`, and `name`, which is fine for a WhatsApp file message (which can include all three properties), but not a Viber file message, which only has `url` and `name`. 

From my understanding of the code, it looks like it would be possible to construct a Viber file message which included a `caption` property (which would then result in error on the API side). Or perhaps I'm missing something in my understanding of `__construct`?

SecondeJK:
@superchilled That's true: you could actually send a file object with the caption for Viber - the way it works is with default values so that it's up to the user to populate the object correctly.

As the channels become more complex, I might introduce a validation set of classes. I don't want to extend the file objects for every channel - there's already more objects than I'd like. But with a validator, it could decorate them to say "you've done it wrong"

superchilled (approved):

superchilled:
> @superchilled That's true: you could actually send a file object with the caption for Viber - the way it works is with default values so that it's up to the user to populate the object correctly.
> 
> As the channels become more complex, I might introduce a validation set of classes. I don't want to extend the file objects for every channel - there's already more objects than I'd like. But with a validator, it could decorate them to say "you've done it wrong"

Thanks Jim. That makes sense! :)

PR #530, opened 2025-01-20
allow php 8.4

ankurk91:
This pull request https://github.com/Vonage/vonage-php-sdk-core/pull/528 is incomplete without this change

Please merge and tag a new minor release


SecondeJK (approved):
Thanks for this, will get a release out shortly

